PR #47, opened 2022-01-02
pin gitlab so upstream mods don't break nebulous init

johndietz:


JaredEdwards (approved):

johndietz:
targeting the 1.6.0 release for this change

PR #70, opened 2022-07-04
Add progress bars

6za:

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -36,6 +36,12 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+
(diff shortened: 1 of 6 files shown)


cmd/create.go
```diff
@@ -48,6 +54,15 @@ This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
 	Run: func(cmd *cobra.Command, args []string) {
 
+		flare.SetupProgress(4)
+		Trackers = make(map[string]*flare.ActionTracker)
```
converge, line 58: is there some benefit of using a pointer in this case?

converge (approved):

PR #81, opened 2022-07-11
feat/cluster-name-and-version-gitops

pagottoo:
Add cluster name to avoid name collision on AWS.
Add version-gitops to clone specific branch when clone gitops repo.


pagottoo:
closed in favor of #82

PR #84, opened 2022-07-11
chore: merge arcocdsync and clean commands

converge:

### DIFF
--- a/cmd/argocdSync.go
+++ b/cmd/argocdSync.go
@@ -1,0 +1,43 @@
+package cmd
+
+import (
+	"log"
+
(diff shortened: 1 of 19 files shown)


6za:
Just a reminder, due to the new `wait` commands this PR code when run `create --dry-run=true`, it is expected long or ethereal waits, as there is not backend to make the conditions true. 

In future, we may want to skip wait situations, on `dry-run` mode. 

Also, some of this will be addressed when added the informers.

jarededwards (approved):

6za (approved):

PR #94, opened 2022-07-13
Adding new shell choices to help on terraform calls

6za:

### DIFF
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -251,30 +251,25 @@ func ApplyGitlabTerraform(dryRun bool, directory string) {
 		}
 		//* AWS_SDK_LOAD_CONFIG=1
 		//* https://registry.terraform.io/providers/hashicorp/aws/2.34.0/docs#shared-credentials-file
-		os.Setenv("AWS_SDK_LOAD_CONFIG", "1")
-		os.Setenv("AWS_PROFILE", config.AwsProfile)
+		envs := map[string]string{}
+		envs["AWS_SDK_LOAD_CONFIG"]="1"
+		envs["AWS_PROFILE"]=config.AwsProfile
 		// Prepare for terraform gitlab execution
-		os.Setenv("GITLAB_TOKEN", viper.GetString("gitlab.token"))
-		os.Setenv("GITLAB_BASE_URL", viper.GetString("gitlab.local.service"))
+		envs["GITLAB_TOKEN"]=viper.GetString("gitlab.token")
+		envs["GITLAB_BASE_URL"]=viper.GetString("gitlab.local.service")
 
 		directory = fmt.Sprintf("%s/.kubefirst/gitops/terraform/gitlab", config.HomePath)
 		err := os.Chdir(directory)
 		if err != nil {
 			log.Panic("error: could not change directory to " + directory)
 		}
-		tfInitCmd := exec.Command(config.TerraformPath, "init")
-		tfInitCmd.Stdout = os.Stdout
-		tfInitCmd.Stderr = os.Stderr
-		err = tfInitCmd.Run()
-		if err != nil {
+		tfInitCmdErr := pkg.ExecShellWithVars(envs,config.TerraformPath, "init")
+		if tfInitCmdErr != nil {
 			log.Panicf("error: terraform init for gitlab failed %s", err)
 		}
 
-		tfApplyCmd := exec.Command(config.TerraformPath, "apply", "-auto-approve")
-		tfApplyCmd.Stdout = os.Stdout
-		tfApplyCmd.Stderr = os.Stderr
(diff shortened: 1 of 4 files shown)


internal/gitlab/gitlab.go
```diff
@@ -251,31 +251,26 @@ func ApplyGitlabTerraform(dryRun bool, directory string) {
...
 		if err != nil {
 			log.Panic("error: could not change directory to " + directory)
 		}
-		tfInitCmd := exec.Command(config.TerraformPath, "init")
-		tfInitCmd.Stdout = os.Stdout
-		tfInitCmd.Stderr = os.Stderr
-		err = tfInitCmd.Run()
-		if err != nil {
-			log.Panicf("error: terraform init for gitlab failed %s", err)
+		tfInitCmdErr := pkg.ExecShellWithVars(envs,config.TerraformPath, "init")
```
converge, line 266: In Go, we usually have `err` as the default variable name, it also helps to have the code more concise.

internal/gitlab/gitlab.go
```diff
@@ -251,31 +251,26 @@ func ApplyGitlabTerraform(dryRun bool, directory string) {
 		}
 		//* AWS_SDK_LOAD_CONFIG=1
 		//* https://registry.terraform.io/providers/hashicorp/aws/2.34.0/docs#shared-credentials-file
-		os.Setenv("AWS_SDK_LOAD_CONFIG", "1")
-		os.Setenv("AWS_PROFILE", config.AwsProfile)
+		envs := map[string]string{}
```
converge, line 254: we could have less variables using `config.EnvVariable`, and AFAIK `viper` is global, we can just call `viper.GetString("value")`

pkg/shell.go
```diff
@@ -20,3 +23,82 @@ func ExecShellReturnStrings(command string, args ...string) (string, string, err
 	log.Println("Commad Execution STDERR: %s", errb.String())
 	return outb.String(), errb.String(), err
 }
+
+func ExecShellReturnStringsWithVars(osvars map[string]string, command string, args ...string) (string, string, error) {
+	
+	log.Printf("INFO: Running %s",command)
+	for k, v := range osvars {
```
converge, line 30: I guess the shell calls are a temporary solution, what do you think to make it explicit the environment variables we're exposing?

pagottoo (approved):
lgtm

internal/gitlab/gitlab.go
6za: I am not aware of a better way to address it, I used the same standard of tools that create HTTP requests with headers.

A map is provided and the framework translates into headers or vars or post body, and so on.

pkg/shell.go
```diff
@@ -20,3 +23,82 @@ func ExecShellReturnStrings(command string, args ...string) (string, string, err
...
+	log.Println("Commad Execution: %s", command)
+	log.Println("Commad Execution STDOUT: %s", outb.String())
+	log.Println("Commad Execution STDERR: %s", errb.String())
+	return outb.String(), errb.String(), err
+}
+
+// Exec shell actions supporting:
+// - On-the-fly logging of result
+// - Map of Vars loaded
+func ExecShellWithVars(osvars map[string]string, command string, args ...string) (error) {
```
converge, line 51: one pattern that I have been trying to follow, and I believe it's nice for developers, is to add a longer description on what the functions is doing, like:
`// ExecShellWithVars does this and that`

it will be beneficial on the long run at the moment we want to export the code documentation to godocs, example: https://pkg.go.dev/io#ByteScanner

pkg/shell.go
```diff
@@ -20,3 +23,82 @@ func ExecShellReturnStrings(command string, args ...string) (string, string, err
...
+	
+	
+	log.Printf("INFO: Running %s",command)
+	for k, v := range osvars {		
+		os.Setenv(k, v)
+		log.Printf(" export %s = %s", k, v)
+	}
+	cmd := exec.Command(command, args...)
+	cmdReaderOut, _ := cmd.StdoutPipe()
```
converge, line 60: it would be nice to treat the errors

converge (approved):
note for later; at some point, it will be nice to update the shell calls that are OS dependent , to use SDK/APIs to call the tools we need like terraform, and curl.

pkg/shell.go
6za: I updated to return all errors to the caller. And log that something happen too.

pkg/shell.go
6za: Not sure what to do yet, I would like to setup some godocs, but there is not much fancy stuff to be done there as fare I found.  This PR is more to improve the lops to not break the progress bars.

6za:
A reminder, this PR is just to help to have less `stdout` of shell commands, to help make easier to improve the ui.

PR #95, opened 2022-07-13
feat: update files and folders to follow rebranding

converge:
this PR:

- update `~/.flare` to `~/.kubefirst`
- update `~/.kubefirst` folder to `~/.k1srt`


cmd/clean.go
```diff
@@ -26,15 +26,25 @@ to quickly create a Cobra application.`,
...
+		}
+
+		err = os.Remove(config.KubefirstConfigFilePath)
+		if err != nil {
+			log.Panicf("unable to delete %q file, error is: ", err)
+		}
 		log.Println("removed $HOME/.kubefirst and $HOME/.flare")
-		if err := os.Mkdir(fmt.Sprintf("%s/.kubefirst", config.HomePath), os.ModePerm); err != nil {
-			log.Panicf("error: could not create directory $HOME/.kubefirst - it must exist to continue %s", err)
+
```
6za, line 39: does this log need to be updated?

cmd/clean.go
```diff
@@ -26,15 +26,25 @@ to quickly create a Cobra application.`,
 
 		log.Println("removing $HOME/.kubefirst and $HOME/.flare")
 		// todo ask for user input to verify?
```
6za, line 28: Maybe should we use the configs? instead of the name in the log string...

6za:
Minor comments on clean step referring to what I guess is the old naming.

cmd/clean.go
converge: nice catch! It was updated

cmd/clean.go
converge: that's true, It was updated

PR #106, opened 2022-07-14
remove more noise from output

6za:


converge (changes requested):
let's try `go fmt` in that single file?

6za:
```bash 
go fmt cmd/createUtils.go
go fmt cmd/destroy.go
go fmt cmd/kubernetes.go
go fmt internal/argocd/argocd.go
go fmt internal/gitlab/gitlab.go
go fmt internal/helm/helm.go
go fmt internal/k8s/kubernetes.go
go fmt internal/vault/vault.go
```

converge (approved):

PR #107, opened 2022-07-14
feat: add handoff for create command

converge:
- adds create handoff screen
- adds docker colors support
- decrease colors to 256 to avoid color mismtach between different terminal apps

Screenshot 2022-07-15 at 12 42 34

notes;
1. our `~.kubefirst` is constantly been updated, this screen version won't have all the data/services.

2. at this point in time, passwords and tokens are shown in the screen, we might want to find a better solution when it gets more mature. maybe encrypting sensitive values.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -8,6 +8,7 @@ import (
 	"github.com/kubefirst/kubefirst/internal/gitlab"
 	"github.com/kubefirst/kubefirst/internal/helm"
 	"github.com/kubefirst/kubefirst/internal/progressPrinter"
+	"github.com/kubefirst/kubefirst/internal/reports"
 	"github.com/kubefirst/kubefirst/internal/softserve"
 	"github.com/kubefirst/kubefirst/internal/terraform"
 	"github.com/kubefirst/kubefirst/internal/vault"
@@ -15,14 +16,10 @@ import (
 	"github.com/spf13/viper"
 	"log"
 	"os/exec"
+	"strings"
 	"syscall"
 	"time"
 )
-
-const trackerStage20 = "0 - Apply Base"
-const trackerStage21 = "1 - Temporary SCM Install"
-const trackerStage22 = "2 - Argo/Final SCM Install"
-const trackerStage23 = "3 - Final Setup"
 
 // createCmd represents the create command
 var createCmd = &cobra.Command{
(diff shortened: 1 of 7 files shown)


cmd/create.go
```diff
@@ -52,7 +48,11 @@ to quickly create a Cobra application.`,
 			log.Panic(err)
 		}
 
+		// todo:
+		// isolate commands, in case we want to run some validations on the create, it would be a good idea to call the
+		// functions that does the validations
 		infoCmd.Run(cmd, args)
+
```
6za, line 55: is the screen locking the terminal?

cmd/create.go
```diff
@@ -259,6 +259,31 @@ to quickly create a Cobra application.`,
...
+			ClusterName:       viper.GetString("cluster-name"),
+			AwsAccountId:      viper.GetString("aws.accountid"),
+			AwsHostedZoneName: viper.GetString("aws.hostedzonename"),
+			AwsRegion:         viper.GetString("aws.region"),
+			ArgoCDUrl:         viper.GetString("argocd.local.service"),
+			ArgoCDUsername:    viper.GetString("argocd.admin.username"),
+			ArgoCDPassword:    viper.GetString("argocd.admin.password"),
+			VaultUrl:          viper.GetString("vault.local.service"),
+			VaultToken:        viper.GetString("vault.token"),
+		}
```
6za, line 279: Do we want show or produce a json or yaml file so people can hook their automations on our post-installs?

internal/reports/create.go
```diff
@@ -0,0 +1,50 @@
...
+	VaultUrl          string
+	VaultToken        string
+}
+
+func BuildCreateHandOffReport(clusterData CreateHandOff) bytes.Buffer {
+
+	var handOffData bytes.Buffer
+	handOffData.WriteString(strings.Repeat("-", 70))
+	handOffData.WriteString(fmt.Sprintf("\nCluster %q is up and running!:\n", clusterData.ClusterName))
+	handOffData.WriteString(strings.Repeat("-", 70))
```
6za, line 26: Should we add a warn to press a button to leave the screen?

internal/reports/create.go
```diff
@@ -0,0 +1,50 @@
...
+	handOffData.WriteString(fmt.Sprintf("\n URL: %s\n", clusterData.ArgoCDUrl))
+	handOffData.WriteString(fmt.Sprintf(" username: %s\n", clusterData.ArgoCDUsername))
+	handOffData.WriteString(fmt.Sprintf(" password: %s\n", clusterData.ArgoCDPassword))
+	handOffData.WriteString(strings.Repeat("-", 70))
+
+	handOffData.WriteString("\n\n--- Vault ")
+	handOffData.WriteString(strings.Repeat("-", 59))
+	handOffData.WriteString(fmt.Sprintf("\n URL: %s\n", clusterData.VaultUrl))
+	handOffData.WriteString(fmt.Sprintf(" token: %s\n", clusterData.VaultToken))
+	handOffData.WriteString(strings.Repeat("-", 70))
```
6za, line 46: Should we clean the terminal window before closing it, to avoid passwords to live on the terminal buffer for too long. In case of live-demos, this may make some devs to want to suppress this screen to avoid conference attacks by people watching the live stream or something.  ;)

cmd/create.go
converge: not anymore, in this PR there is an update to use a non blocker call.

cmd/create.go
converge: kind of.. that was my second point in the PR notes above. we can do it, but it would be better to provide some encryption layer to avoid plain text with sensitive data.

internal/reports/create.go
converge: IMHO the user level that we are aiming know that ctrl+c will quit

internal/reports/create.go
converge: I agree, but I would like to propose a different approach:

idea a) ask the user like "hey, we're going to provide some sensitive data, do you want to see it now? (yes/no)

idea b) tell the user that the sensitive data is stored at the place `.../folder/file` and he can use a decryption key to open it

the current solution is something quick for us to start, we can keep improving it.

PR #111, opened 2022-07-15
chore: update info screen to be non blocking screen

converge:


6za (approved):
I trust is good, not familiar with the change.

PR #123, opened 2022-07-19
fix: remove required env value, and allow init command to collect it

converge:

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ The setup is extremely simple, create a `.env` file in the root folder, and add 
 | Variable           | example          |
 |--------------------|------------------|
 | AWS_PROFILE        | default          |
-| AWS_REGION         | us-east-1        |
 | CLOUD_PROVIDER=aws | aws              |
 | HOSTED_ZONE_NAME   | example.com      |
 | ADMIN_EMAIL        | john@example.com |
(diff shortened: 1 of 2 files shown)


pagottoo (dismissed):
lgtm

cmd/init.go
```diff
@@ -80,6 +81,11 @@ to quickly create a Cobra application.`,
 		// name of the cloud region to provision resources when resources are region-specific
 		region, _ := cmd.Flags().GetString("region")
 		viper.Set("aws.region", region)
+		// propagate it to local environment
+		err = os.Setenv("AWS_REGION", region)
+		if err != nil {
+			log.Panicf("unable to set environment variable AWS_REGION, error is: %v", err)
+		}
```
6za, line 88: This need to be a required flag, not a variable. Logs is also not reflecting it.

6za: `os.Setenv` due to some weird ways go fork shells, doesn't produce the expected behavior, specially for terraform.

6za:
I believe, that terraforms all reading from the viper configs, I would doublecheck, but I think we are good.

cmd/init.go
converge: it was actually there:
```
	initCmd.Flags().String("region", "", "the region to provision the cloud resources in")
	err = initCmd.MarkFlagRequired("region")
	if err != nil {
		log.Panic(err)
	}
```

6za: As this is a flag, the message should refer to a missing flag, or a wrong valuer of a flag. Not a variable, variables means a "Environment Variable" on this context for most users.

6za: Ahh, I see.. it is a error setting.. never mind. Approving.

6za (approved):

PR #135, opened 2022-07-19
feat-restore-certs

pagottoo:
Download SSL/TLS files from S3 and apply on k8s cluster to avoid re-issuer certificates.


cmd/restoreSSL.go
```diff
@@ -0,0 +1,42 @@
+/*
```
converge, line 1: it would be nice to remove left overs

cmd/restoreSSL.go
```diff
@@ -0,0 +1,42 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("restoreSSL called")
+		ssl.RestoreSSL()
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(restoreSSLCmd)
+
+	// Here you will define your flags and configuration settings.
```
converge, line 33: it would be nice to remove left overs

internal/aws/aws.go
```diff
@@ -339,3 +341,106 @@ func DestroyBucketsInUse(destroyBuckets bool) {
...
+	downloader := s3manager.NewDownloader(GetAWSSession())
+
+	log.Println("Listing the objects in the bucket:")
+	listObjsResponse, err := s3Client.ListObjectsV2(&s3.ListObjectsV2Input{
+		Bucket: aws.String(bucket),
+		Prefix: aws.String(""),
+	})
+
+	if err != nil {
+		panic("Couldn't list bucket contents")
```
converge, line 424: would be nice to log it

internal/ssl/ssl.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/kubefirst/kubefirst/internal/aws"
+	"github.com/kubefirst/kubefirst/internal/k8s"
+	"github.com/kubefirst/kubefirst/pkg"
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var namespaces []string
```
converge, line 22: it would be a good idea to remove globals, we can host it inside Configs

internal/aws/aws.go
```diff
@@ -330,3 +332,107 @@ func DestroyBucketsInUse(destroyBuckets bool) {
...
+	} else {
+		log.Printf("[#99] Dry-run mode, bucket creation skipped:  %s", name)
+	}
+	viper.Set(fmt.Sprintf("bucket.%s.created", name), true)
+	viper.Set(fmt.Sprintf("bucket.%s.name", name), name)
+	viper.WriteConfig()
+}
+
+func UploadFile(bucket, key, fileName string) error {
+	// The session the S3 Uploader will use
```
converge, line 375: *left over

internal/aws/aws.go
```diff
@@ -330,3 +332,107 @@ func DestroyBucketsInUse(destroyBuckets bool) {
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to upload file, %v", err)
+	}
+	log.Printf("file uploaded to, %s\n", result.Location)
+	return nil
+}
+
+func DownloadBucket(bucket string, destFolder string) error {
+	// Create a file to write the S3 Object contents to.
```
converge, line 400: *left over

internal/ssl/ssl.go
```diff
@@ -0,0 +1,201 @@
...
+				return nil, err
+			}
+			files = append(files, fullFileName)
+		}
+	}
+
+	return files, nil
+}
+
+//func GetBackupCertificates(apiGroup string, apiVersion string, resourceTypes []string, namespace string) ([]string, error) {
```
converge, line 85: *left over

internal/ssl/ssl.go
pagottoo: resolved using a func to return the list of namespaces

converge (approved):

johndietz:
thanks for getting this in today guys

PR #149, opened 2022-07-22
Retry certs logic

6za:
Exploring ways to retry cert for HTTPS on ArgoCD


converge (approved):
in the `cmd`, instead of log.Panic, we can use `log.Println() ; return` on most of the cases.

pagottoo (approved):
lgtm

converge (approved):

PR #150, opened 2022-07-25
put-tag-bucket-delete-on-clean

pagottoo:
With this PR we have 2 features:

- Put tags on buckets (`provisioned-by` and `cluster-name`);
- During `clean` command we have a new flag (`--destroy-buckets`) to remove buckets from cloud provisioned when `init` command was executed.
- We ask from user to confirm the destroy operation answering a question, and introduce another flag (`--destroy-confirm`) to work together  `--destroy-buckets` that skip the question (and can be used in automation scripts)


cmd/clean.go
```diff
@@ -3,28 +3,47 @@ package cmd
...
-re-create all Kubefirst files.`,
+	Long: `Kubefirst creates files, folders and cloud buckets during installation at your environment. This command removes and 
+re-create all Kubefirst files. To destroy cloud resources you need to specify aditional flags (--destroy-buckets)`,
 	Run: func(cmd *cobra.Command, args []string) {
 
 		config := configs.ReadConfig()
 
+		destroyBuckets, err := cmd.Flags().GetBool("destroy-buckets")
+		if err != nil {
+			log.Panic(err)
```
converge, line 32: in this case we can use `log.Println()` and call return, this will avoid printing the stacktrace to the console.

cmd/clean.go
```diff
@@ -3,28 +3,47 @@ package cmd
...
+		}
+		destroyConfirm, err := cmd.Flags().GetBool("destroy-confirm")
+		if err != nil {
+			log.Panic(err)
+		}
+		if destroyBuckets && !destroyConfirm {
+			destroyConfirm = pkg.AskForConfirmation("This process will delete cloud buckets and all files inside, do you really want to proceed?")
+			if !destroyConfirm {
+				os.Exit(130)
```
converge, line 41: out of curiosity, what does the 130 means?

6za: https://www.linuxdoc.org/LDP/abs/html/exitcodes.html

6za: 130 | script terminated by Control-C
-- | --

6za: User desired to abort execution.

cmd/clean.go
pagottoo: done

6za (approved):

internal/aws/aws.go
```diff
@@ -427,3 +422,108 @@ func DownloadBucket(bucket string, destFolder string) error {
...
+				{
+					Key:   aws.String("ClusterName"),
+					Value: aws.String(clusterName),
+				},
+			},
+		},
+	}
+
+	_, err := svc.PutBucketTagging(input)
+	if err != nil {
```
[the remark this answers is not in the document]
pagottoo: done

converge (approved):

PR #152, opened 2022-07-26
fix: remove unecessary argo cd sync call

converge:

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -144,7 +144,7 @@ to quickly create a Cobra application.`,
 		progressPrinter.IncrementTracker("step-argo", 1)
 
 		informUser("Getting an argocd auth token")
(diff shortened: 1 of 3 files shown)


cmd/create.go
```diff
@@ -153,34 +153,7 @@ to quickly create a Cobra application.`,
 			}
 			time.Sleep(45 * time.Second)
 		}
-
-		informUser("Syncing the registry application")
```
6za, line 157: I am not 100% sure we can remove this.. 
In some installs, we got in a state that this synch was needed, due to some softserve/argo miscommunication.

I believe we should not block, the installer in case is already happening, but not 100% on remove it to ensure consistency.

converge: I had 2 successful runs for different regions using this PR.

johndietz (approved):

6za (approved):

PR #153, opened 2022-07-26
Destroy improvements

6za:
- Makes Telemetry not panic
- Add some UI to Destroy action
- Add Dry-run mode to Destroy

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -6,11 +6,13 @@ import (
 	"log"
 	"os/exec"
 	"syscall"
+	"time"
 
 	"github.com/kubefirst/kubefirst/configs"
 	"github.com/kubefirst/kubefirst/internal/gitlab"
 	"github.com/kubefirst/kubefirst/internal/k8s"
 	"github.com/kubefirst/kubefirst/internal/terraform"
+	"github.com/kubefirst/kubefirst/internal/progressPrinter"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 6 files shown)


pagottoo (dismissed):
lgtm

cmd/destroy.go
```diff
@@ -39,6 +41,17 @@ if the registry has already been deleted.`,
 		if err != nil {
 			log.Panic(err)
 		}
+		dryRun, err := cmd.Flags().GetBool("dry-run")
+		if err != nil {
+			log.Panic(err)
```
converge, line 46: it will avoid stack trace / and will return the error message to the user
```suggestion
			log.Println(err)
			return
```

converge (dismissed):

cmd/root.go
```diff
@@ -15,7 +14,9 @@ var rootCmd = &cobra.Command{
 	open source application delivery platform in under an hour. 
 	checkout the docs at docs.kubefirst.com.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		log.Println(viper.Get("name"))
+		//log.Println(viper.Get("name"))
```
converge, line 17:
```suggestion
```

converge (approved):

cmd/destroy.go
6za: I am trying to following the existing pattern, not refactor anything at the moment.

cmd/root.go
6za: Is this a blocker?

pagottoo (approved):
lgtm

PR #169, opened 2022-07-28
volume mount for cloud credentials

jarededwards:


converge:
tks Jared, we included some other updates for this issue here https://github.com/kubefirst/kubefirst/pull/170

feel free to re-open the PR if there is something else we should add

PR #184, opened 2022-08-03
Feat remove env vars

### LINKED ISSUE #179: remove dependency on environment variables
we should add default cli flags that can accommodate this so we dont need to explicitly set envs up front
AWS_PROFILE
AWS_REGION

converge:
fix #179


cmd/init.go
```diff
@@ -97,6 +99,15 @@ to quickly create a Cobra application.`,
 		}
 		log.Println("region:", region)
 
+		profile, _ := cmd.Flags().GetString("profile")
+		viper.Set("aws.profile", profile)
```
6za, line 103: Should we be checking for error here? as the other ones..

6za: I saw that is later..

cmd/init.go
```diff
@@ -97,6 +99,15 @@ to quickly create a Cobra application.`,
...
+		// propagate it to local environment
```
6za, line 104: Should we check for error here?
profile, _ := cmd.Flags().GetString("profile")

```golang
profile, err := cmd.Flags().GetString("profile")
		if err != nil {
			log.Println(err)
		}
```

converge: its always nice to be ready the day it will go wrong xD

pagottoo (approved):
/lgtm

Issue #188, opened 2022-08-04
Error Processing  gitlab.ChangeRegistryToGitLab   - empty secret


6za:
```bash
panic: error creating argocd repository credentials template the object provided is unrecognized (must be of type Secret): couldn't get version/kind; json parse error: unexpected end of JSON input (<empty>)

goroutine 1 [running]:
log.Panicf({0x1ccfa77, 0x0}, {0xc0009874f8, 0x0, 0x0})
  /usr/local/go/src/log/log.go:361 +0x67
github.com/kubefirst/kubefirst/internal/gitlab.ChangeRegistryToGitLab(0x0)
  /home/developer/app/internal/gitlab/gitlab.go:430 +0x58e
github.com/kubefirst/kubefirst/cmd.glob..func6(0x2f6a100, {0x2fb7958, 0x0, 0x0})
  /home/developer/app/cmd/create.go:297 +0x2065
github.com/spf13/cobra.(*Command).execute(0x2f6a100, {0x2fb7958, 0x0, 0x0})
  /go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:860 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0x2f6a380)
  /go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3bc
github.com/spf13/cobra.(*Command).Execute(...)
  /go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
github.com/kubefirst/kubefirst/cmd.Execute()
  /home/developer/app/cmd/root.go:26 +0x25
main.main()
  /home/developer/app/main.go:65 +0x425
```

# Logs

```bash
LOG: 2022/08/04 00:12:19.362648 /home/developer/app/internal/gitlab/gitlab.go:638: successfully pushed metaphor to gitlab
LOG: 2022/08/04 00:12:20.378060 /home/developer/app/cmd/createUtils.go:295: Adding the registry application registered against gitlab
LOG: 2022/08/04 00:12:20.381526 /home/developer/app/internal/gitlab/gitlab.go:425: error unmarshalling yaml during argocd repository secret create error converting YAML to JSON: yaml: line 2: found character that cannot start any token
[...]

PR #190, opened 2022-08-04
#188  2nd  try - fix tabs

6za:
#188 - Second attempt to fix it

### DIFF
--- a/internal/gitlab/gitlab.go
+++ b/internal/gitlab/gitlab.go
@@ -422,7 +422,6 @@ func ChangeRegistryToGitLab(dryRun bool) {
 			Type: "Opaque",
 		}
 
-		log.Println(argocdRepositoryAccessTokenSecret.String())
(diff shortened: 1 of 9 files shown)


internal/gitlab/gitlab.go
```diff
@@ -440,14 +443,17 @@ func ChangeRegistryToGitLab(dryRun bool) {
...
+			annotations:
+				managed-by: argocd.argoproj.io
+			labels:
+				argocd.argoproj.io/secret-type: repository
+			name: repo-gitlab
+			namespace: argocd
 		type: Opaque
-	`)
+		`)
+		if err != nil {
```
converge, line 454: in case `err is nil` , next condition will never happen

6za: I believe thata is the idea, no?
- If the template don't work, we can execute it application.. 
- Am I handling the same error twice? is this your concern?

pagottoo (approved):

PR #221, opened 2022-08-12
Add Github Initial Effort

6za:
- Adding github support to kubefirst
- Add github related actions to help test github integration


6za:
I would like to handle: https://github.com/kubefirst/kubefirst/issues/222

Issue #233, opened 2022-08-16
Add CLI version to telemetry data


6za:
Allow telemetry reports to know version used for install runs.

PR #234, opened 2022-08-16
Solve #233

6za:
Adds cli version to telemetry


6za:
On hold, waiting metric collector to be updated to process change.

6za:
now, metrics flowing as expected.

converge:
it would be nice to receive config as a parameter, it avoid instantiating a new config, and is future proof for the methods implementation that will come on a later point.

pagottoo (approved):

PR #249, opened 2022-08-17
chore: add timestamp to s3 file

converge:


6za (approved):

converge:
as discussed with the team, we're using S3 versioning in favor of adding timestamps to the file

PR #250, opened 2022-08-17
docker image to run kubefirst bin inside container 

pagottoo:
This PR introduces a docker image (the minimal possible), to run kubefirst binary inside a container (to support other OS, like Windows). 

Signed-off-by: Thiago Pagotto <pagottoo@gmail.com>


pagottoo:
@johndietz @jarededwards @6za @converge @CristhianF7 this is an initial idea and I would like to count on your review.

README.md
```diff
@@ -61,33 +61,23 @@ gitlab.subdomain.example.com
...
-git clone https://github.com/kubefirst/kubefirst.git
-
-# via SSH
-git clone git@github.com:kubefirst/kubefirst.git
-```
+#Check the release page:
+#https://github.com/kubefirst/kubefirst/releases
 
-## Start the Container
+export KUBEFIRST_VERSION=1.8.5
```
johndietz, line 70: i think we may want this to read 1.8.6 and update the config.Version to match for the 1.8.6 publish operation to align after merge

build/README.md
```diff
@@ -0,0 +1,39 @@
+## Runing inside docker
+
+```bash
+docker run \
+  -it --name kubefirst  \
+  --dns="1.0.0.1" --dns="208.67.222.222" --dns="8.8.8.8" \
+  -v $(PWD):/opt/kubefirst \
+  -v $HOME/.aws:/home/developer/.aws \
+   pagottoo/kubefirst_cli:1.8.5
```
johndietz, line 9: should we make this
```suggestion
   kubefirst/kubefirst:1.8.6
```
and i'll publish 1.8.6 to dockerhub as part of the release?

build/Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM --platform=linux/amd64 golang:1.18
+FROM --platform=linux/amd64 buildpack-deps:bullseye-scm
+# we are using buidlpack-deps:bullseye-scm https://github.com/docker-library/golang/blob/8d0fa6028120904e16fe761f095bd0620b68eab2/1.18/bullseye/Dockerfile
 
-WORKDIR /opt/kubefirst-install
+ARG KUBEFIRST_VERSION=1.8.5
```
johndietz, line 4: same 1.8.6 comment here - i may be mistaken with the approach, or i may also just early reviewing the pr and this is set up so you can test these changes

docker-compose.yaml
```diff
@@ -6,7 +6,7 @@ services:
     platform: linux/amd64
     build:
       context: .
-      dockerfile: ./build/Dockerfile
```
johndietz, line 9: if the users are going to leverage the readme for a container build process, and build/Dockerfile is the one that's pinned to a version, i think this should stay build/Dockerfile right? let's sync on all the pieces in the morning so i'm sure i know how the story works.

6za: As discussed, this will the user community version on `./build`. Thanks for the input.

build/README.md
6za: yep, thanks.

build/Dockerfile
6za: updated, thanks.

johndietz (approved):

PR #261, opened 2022-08-22
feat: add go routines to init command

converge:

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -31,10 +31,7 @@ Cobra is a CLI library for Go that empowers applications.
 This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		/**
-		BLOCK 1
-		*/
-		startTime := time.Now()
+
 		config := configs.ReadConfig()
 		infoCmd.Run(cmd, args)
 
@@ -227,8 +224,6 @@ to quickly create a Cobra application.`,
 		//! tracker 8
 		progressPrinter.IncrementTracker("step-telemetry", 1)
 		time.Sleep(time.Millisecond * 100)
-
-		fmt.Printf("total init command elapsed time: %.2f", time.Since(startTime).Seconds())
 
 		return nil
 	},
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/kubefirst/kubefirst/configs"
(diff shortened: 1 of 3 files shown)


cmd/init.go
```diff
@@ -54,126 +72,144 @@ to quickly create a Cobra application.`,
...
+
+			if len(awsFlags.AssumeRole) > 0 {
+				log.Println("calling assume role")
+				err := aws.AssumeRole(awsFlags.AssumeRole)
+				if err != nil {
+					log.Println(err)
+					//return err
+				}
+				log.Printf("assuming new AWS credentials based on role %q", awsFlags.AssumeRole)
 			}
```
6za, line 88: I believe for safety, this step should be kept serial.. as I believe the bucket creation step will use this information to create a bucket with the role in play.

6za:
Reviewing the PR for possible race conditions.. as this may impact installer behavior.

internal/downloadManager/download.go
```diff
@@ -30,116 +31,135 @@ func DownloadTools(config *configs.Config) error {
...
-	extractFileFromTarGz(
-		helmTarDownload,
-		fmt.Sprintf("%s-%s/helm", config.LocalOs, config.LocalArchitecture),
-		config.HelmClientPath,
-	)
-	err = os.Chmod(config.HelmClientPath, 0755)
-	if err != nil {
-		return err
-	}
+	go func() {
```
6za, line 112: Maybe we should use named functions here, to make tis code better to read, also cleaner. 

This function is the one most improved by been parallelized.

cmd/init.go
```diff
@@ -54,126 +72,144 @@ to quickly create a Cobra application.`,
...
+
+		go func() {
+			//! tracker 7
+			log.Printf("cloning and detokenizing the metaphor-template repository")
+			prepareKubefirstTemplateRepo(config, viper.GetString("gitops.owner"), "metaphor", "", viper.GetString("template.tag"))
+			log.Println("clone and detokenization of metaphor-template repository complete")
+			progressPrinter.IncrementTracker("step-metaphor", 1)
+
+			wg.Done()
+		}()
```
6za, line 210: This step was removed in current iterations of init in construction.

cmd/init.go
```diff
@@ -54,126 +72,144 @@ to quickly create a Cobra application.`,
...
+
+		wg.Wait()
```
6za, line 212: Did you checked if `viper.WriteConfig()` and viper is thread safe? As we have several thread updating values on it. Do we need any mutex protection?

6za: Maybe this could help for details: https://github.com/spf13/viper/issues/19

6za:
I really like the direction of this PR, I just want we ensure we are aware, that we need to ensure:
- Viper is thread-safe to allow the parallel steps to no break as create is heavily dependent of `init`final kubefirst file. 
- Code is kept easy to read, thread(and similar like go routines, add some complexities to manage "order")

converge:
as mentioned https://github.com/spf13/viper/issues/19 , viper is thread unsafe. I'll rewrite the PR at some point to speed up init.

> Code is kept easy to read, thread(and similar like go routines, add some complexities to manage "order")

as you guys are changing a lot of the structure, I don't want to add new functions at this point. I would might need to refactor a bit, and this will cause merging conflicts on you guys current PRs.

Issue #275, opened 2022-08-24
1.9.x - Feature parity - Add Metaphors(go and frontend)


6za:
# Overview

Github is missing frontend/go metaphor

# Goal

To create a common process do deploy metaphor and family to github and gitlab.

6za:
solved on #253

PR #276, opened 2022-08-24
Unify install of metaphor #275

6za:
Add Metaphors app to github.

- #275 

### DIFF
--- a/cmd/deployMetaphor.go
+++ b/cmd/deployMetaphor.go
@@ -6,11 +6,9 @@ package cmd
 
 import (
 	"fmt"
-	"log"
 
-	"github.com/kubefirst/kubefirst/configs"
 	"github.com/kubefirst/kubefirst/internal/flagset"
-	"github.com/kubefirst/kubefirst/internal/gitlab"
+	"github.com/kubefirst/kubefirst/internal/metaphor"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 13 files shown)


6za:
create-action

cmd/githubRemoveValidate.go
```diff
@@ -33,6 +33,14 @@ var githubRemoveValidateCmd = &cobra.Command{
 		if err == nil {
 			fmt.Println("gitops not found as expected")
 		}
+		repoMetaphorGo, err := gitWrapper.GetRepo(owner, "metaphor-go")
+		if err == nil {
```
converge, line 37: is this logic correct? I believe it should be `if err != nil {`

cmd/githubRemoveValidate.go
```diff
@@ -33,6 +33,14 @@ var githubRemoveValidateCmd = &cobra.Command{
...
+			fmt.Println("gitops not found as expected")
+		}
+		repoMetaphorFrontend, err := gitWrapper.GetRepo(owner, "metaphor-frontend")
+		if err == nil {
```
converge, line 41: is this logic correct? I believe it should be if err != nil {

cmd/githubRemoveValidate.go
```diff
@@ -41,7 +49,17 @@ var githubRemoveValidateCmd = &cobra.Command{
 
 		if repoMetaphor.GetName() == "metaphor" {
 			fmt.Println("metaphor should be not present")
-			return fmt.Errorf("error validating repo: %s ", repoGitops.GetName())
+			return fmt.Errorf("error validating repo: %s ", repoMetaphor.GetName())
+		}
+
+		if repoMetaphorGo.GetName() == "metaphor-go" {
```
converge, line 55: can `repoMetaphorGo` be another repo. name?

cmd/githubRemoveValidate.go
6za: yes, this is validate the non-existence of something. If error it is fine.. 

As the next line says: 
```
fmt.Println("gitops not found as expected")
```

6za: it is expected to fail on the validation.

cmd/githubRemoveValidate.go
6za: Same as above.

cmd/githubRemoveValidate.go
6za: ? Not sure of the question, this is just integration tests, not sure the impact of the name.

cmd/deployMetaphor.go
```diff
@@ -22,48 +20,17 @@ var deployMetaphorCmd = &cobra.Command{
...
-		log.Println("clone and detokenization of metaphor-frontend-template repository complete")
-
-		if !viper.GetBool("gitlab.metaphor-pushed") {
-			log.Println("Pushing metaphor repo to origin gitlab")
-			gitlab.PushGitRepo(globalFlags.DryRun, config, "gitlab", "metaphor")
-			viper.Set("gitlab.metaphor-pushed", true)
-			viper.WriteConfig()
-			log.Println("clone and detokenization of metaphor-frontend-template repository complete")
+		if viper.GetBool("github.enabled") {
+			return metaphor.DeployMetaphorGithub(globalFlags)
```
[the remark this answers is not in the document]
6za: Where?
This functions return error, if errors out they will send the error upstream as desired on this code. 

I am using `runE` to allow the error to bubble up to create and the error is handled there.

6za:
Metaphor deployment is registred as expected: 
```
--- Metaphor Frontend-------------------------------------------------
...
----------------------------------------------------------------------
--- Metaphor Go-------------------------------------------------------
...
----------------------------------------------------------------------
--- Metaphor ---------------------------------------------------------
   ...
----------------------------------------------------------------------
```

6za:
image

cmd/githubRemoveValidate.go
6za: it could be empty, so the test it is to ensure it is empty. 
```
fmt.Println("metaphor should be not present") 
```

cmd/deployMetaphor.go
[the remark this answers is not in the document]
converge: my bad, we usually do `if err ...`, youre just returning it to `runE`.

you can do:
```
		if viper.GetBool("github.enabled") {
			return metaphor.DeployMetaphorGithub(globalFlags)
		}
		return metaphor.DeployMetaphorGitlab(globalFlags)
```

then else statement is not necessary

cmd/githubRemoveValidate.go
converge: we usually separate what is application code, and what is test.

it sounds strange to expect something different coming from `repoMetaphorGo.GetName` that is not the application name.

pagottoo (approved):
lgtm

converge (approved):

PR #302, opened 2022-08-30
feat: add silent mode

converge:

### DIFF
--- a/cmd/createGithub.go
+++ b/cmd/createGithub.go
@@ -1,6 +1,5 @@
 /*
 Copyright © 2022 NAME HERE <EMAIL ADDRESS>
-
 */
 package cmd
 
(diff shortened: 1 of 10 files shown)


cmd/createGithub.go
```diff
@@ -53,29 +42,39 @@ var createGithubCmd = &cobra.Command{
...
+		informUser("Creating gitops/metaphor repos", false)
 		err = githubAddCmd.RunE(cmd, args)
 		if err != nil {
 			log.Println("Error running:", githubAddCmd.Name())
 			return err
 		}
-		informUser("Create Github Repos")
+		informUser("Create Github Repos", false)
 		progressPrinter.IncrementTracker("step-github", 1)
 		err = loadTemplateCmd.RunE(cmd, args)
```
6za, line 68: if we use a viper flag, you don't need to change the signature.

converge: I would like to avoid adding more content to the kubefirst file , and having silent mode flag in the viper file doesn't sound good IMHO

marinhojess (approved):

converge:
if there is some points to be discussed, we can continue here. Im merging to unblock the pipeline tests.

Issue #277, opened 2022-08-24
1.9.x - Support  KUBEFIRST_AWS_NODES_SPOT=true


6za:
# Overview

To allow the flag `--aws-nodes-spot` be also set by env variable:

Example:
```bash 
export KUBEFIRST_AWS_NODES_SPOT=true
```

6za:
We would follow `atlantis` style flag control: https://www.runatlantis.io/docs/server-configuration.html#environment-variables

- `KUBEFIRST_*`

6za:
```bash
All flags can be specified as environment variables.

Take the flag name, ex. --aws-nodes-spot
Ignore the first -- => aws-nodes-spot
Convert the -'s to _'s => aws-nodes-spot
Uppercase all the letters => AWS_NODES_SPOT
Prefix with KUBEFIRST_ => KUBEFIRST_AWS_NODES_SPOT
```

6za:
```bash 
export KUBEFIRST_AWS_NODES_SPOT=true
./kubefirst init  --profile default  --cloud aws  --region us-east-1 ...
```


```bash
LOG: 2022/09/01 11:32:36.509606 /home/developer/app/internal/flagset/common.go:54: Enviroment Variable(KUBEFIRST_AWS_NODES_SPOT) set - using this value for Flag(aws-nodes-spot)
LOG: 2022/09/01 11:32:36.509816 /home/developer/app/internal/flagset/aws.go:72: aws.nodes_spot:  true
```

6za:
PR opened.

PR #311, opened 2022-08-31
Adding support for #277 

6za:
- Adding support for #277 
- Adding support to flags or env variables 
- ~Adding some test scenarios~

```bash 
=== RUN   Test_GetFlagVarName
--- PASS: Test_GetFlagVarName (0.00s)
=== RUN   Test_DefineSource_set_by_flag
2022/08/31 19:18:48 Flag(sample) set from CLI flag
--- PASS: Test_DefineSource_set_by_flag (0.00s)
=== RUN   Test_DefineSource_set_by_var
Enviroment Variable(KUBEFIRST_SAMPLE) set - using this value for Flag(sample)
--- PASS: Test_DefineSource_set_by_var (0.00s)
=== RUN   Test_DefineSource_notSet
Enviroment Variable(sample) not set
--- PASS: Test_DefineSource_notSet (0.00s)
=== RUN   Test_DefineSource_set_by_flag_bool
2022/08/31 19:18:48 Flag(sample) set from CLI flag
--- PASS: Test_DefineSource_set_by_flag_bool (0.00s)
PASS
ok  	github.com/kubefirst/kubefirst/internal/flagset	(cached)
```


converge:
what do you think about this reference? https://kubeshop.atlassian.net/wiki/spaces/KUBESHOP/pages/607551493/Tests

the unit tests are touching the real resource, could you mock it to avoid external calls? as an example, when we unit test some function that writes to a file, it's nice to have a mock file writer, then real files are not created.

another example are external http calls, it's a nice when we dont reach the external/3rd party url to run unit tests, to avoid that every time the test run, it request the 3rd party url, but calls our mock to provide the expected response.

pls let me know your thoughts, we can keep improving the docs together.

6za:
Adding confirmation of success: 
```bash 
export KUBEFIRST_AWS_NODES_SPOT=true
./kubefirst init  --profile default  --cloud aws  --region us-east-1 ...
```


```bash
LOG: 2022/09/01 11:32:36.509606 /home/developer/app/internal/flagset/common.go:54: Enviroment Variable(KUBEFIRST_AWS_NODES_SPOT) set - using this value for Flag(aws-nodes-spot)
LOG: 2022/09/01 11:32:36.509816 /home/developer/app/internal/flagset/aws.go:72: aws.nodes_spot:  true
```

6za:
> what do you think about this reference? https://kubeshop.atlassian.net/wiki/spaces/KUBESHOP/pages/607551493/Tests
> 
> the unit tests are touching the real resource, could you mock it to avoid external calls? as an example, when we unit test some function that writes to a file, it's nice to have a mock file writer, then real files are not created.
> 
> another example are external http calls, it's a nice when we dont reach the external/3rd party url to run unit tests, to avoid that every time the test run, it request the 3rd party url, but calls our mock to provide the expected response.
> 
> pls let me know your thoughts, we can keep improving the docs together.

I read it, I tried to follow what was possible, but I found a reference that helped me test, what I needed that was to ensure the command did, what I desired to do. Test we done to support development, if we don't want this on this repo, I can move for a fork, until I am able to comply to the enforced standard.

6za:
Test we removed was it seems they don't conform with some standard.

converge (approved):
I didn't understand what the code is doing to be honest, the `configs/` also loads environment variables, but I dont want to block you.

PR #313, opened 2022-09-01
chore: remove unused variables, and move kubernetes functions to kube…

converge:
- remove unused code
- move kubernetes functions to kubernetes `internal/`


internal/k8s/kubernetes.go
```diff
@@ -49,12 +39,6 @@ func GetPodNameByLabel(podsClient coreV1Types.PodInterface, label string) string
 	return gitlabToolboxPodName
 }
 
-func DeletePodByName(podsClient coreV1Types.PodInterface, podName string) {
-	err := podsClient.Delete(context.TODO(), podName, metaV1.DeleteOptions{})
-	if err != nil {
-		log.Println(err)
-	}
-}
```
6za, line 57: Why removing it?

internal/k8s/kubernetes.go
```diff
@@ -64,34 +48,6 @@ func DeletePodByLabel(podsClient coreV1Types.PodInterface, label string) {
...
-// }
-
-// func CreateCredentialsTemplateSecret() {
-
-// }
-
-func getVaultRootToken(vaultSecretClient coreV1Types.SecretInterface) string {
-	name := "vault-unseal-keys"
-	log.Printf("Reading secret %s\n", name)
-	secret, err := vaultSecretClient.Get(context.TODO(), name, metaV1.GetOptions{})
```
6za, line 78: Does this function really annoy us enough by existing? 

> maybe we can keep for a bit for reference code. 

> It is a nice example of marshal and secret manipulation.

internal/k8s/kubernetes.go
6za: It is a nice utility that I see value downstream, if we stoped using today.

converge: the function is not in use, there is no other part of the code that needs it

internal/k8s/kubernetes.go
converge: the function is not in use, there is no other part of the code that needs it. 

it will keep in git history in case we need it.

YAGNI is a good reference for that

6za (approved):
I am approving this PR with reservations, I would prefer we avoid removing current utilities until we define the goal of our packages on internal.

Some of us see, as we are building up some utilities to be able to integrate more scenarios of installations, if this become a trend that once a function get not used, without been replaced by another one, we just remove it on our `internal` modules. 

For me, this practice may slow down experimentation and tool sharing between members of the team. I agree refractories are important, I liked most of the reorganization done on this PR, I just don't feel we should apply "new rules" without communicating with the team the direction and without some consensus. 

A more appropriated moment is post-release to trigger this "clean task", not just before closing one. 

Consulting history to find code, it is slower process. I think we can find a better midpoint on this. We are building some stuff today, to use tomorrow.

internal/k8s/kubernetes.go
6za: Approved with reservations, prefer we hold the deletion of artifacts.

internal/k8s/kubernetes.go
6za: Approved with reservations, prefer we hold the deletion of artifacts. Details on approval.

Issue #320, opened 2022-09-02
Console  UI - CMD


6za:
# Overview

To support the local UI(#298) we need to embed a API to trigger some actions and collect some info from our viper file. 

# Goal 

- Define a OpenAPI spec of that API. 
- Add a route to read viper configs, generic route for all properties. 
- Add a route to get the state/status of a given domain(not resolving, resolving, certs not good, certs good, returns as expected, and other signals.)

PR #336, opened 2022-09-08
Console and Internal API implementation

CristhianF7:
Fixes: https://github.com/kubefirst/kubefirst/issues/320

- Adding a new console command

1. Kubefirst console.
2. It downloads the latest version dist from GitHub
3. Starts the console app in port 8080 using that dist folder

- Adding a new API command

1. It reads the config values from viper
2. Servers an endpoint running on 8080/configs


cmd/console.go
```diff
@@ -0,0 +1,32 @@
...
+	Long:  "starts app server for the Kubefirst console",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		config := configs.ReadConfig()
+		distFolder := fmt.Sprintf("%s/tools/console/dist", config.K1FolderPath)
+		fileServer := http.FileServer(http.Dir(distFolder))
+		http.Handle("/", fileServer)
+
+		fmt.Printf("Starting server at port 8080\n")
+	
+		if err := http.ListenAndServe(":8080", nil); err != nil {
```
converge, line 23: shall we use a less common port? it would avoid port conflict in case the user has some local service running in the most used http ports for local access.

many services make use of 8080 for running their service locally, we could have something like 8xxx where xxx is the port you choose

the http port could be a constant, so we dont need to repeat the value, or in case we need to change it, it a single point to change.

build/README.md
```diff
@@ -71,3 +71,14 @@ docker exec -it kubefirst bash
 ```
 
 After this step is executed, return to [this step](https://github.com/kubefirst/kubefirst#initialization) to run a `kubefirst info` and other functions.
+
+
+### Running CLI in docker container using DEV
```
converge, line 76: that's awesome that you took the time to update the docs!

6za (dismissed):
lgtm, some details it may be addressed later. Great Work!

cmd/console.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/kubefirst/kubefirst/configs"
+	"github.com/spf13/cobra"
+	"log"
+	"net/http"
+)
+
+var console = &cobra.Command{
+	Use:   "console",
+	Short: "starts ui server",
+	Long:  "starts app server for the Kubefirst console",
```
converge, line 14: maybe we could extend the long description a little more, maybe adding that the UI will contain some links, and..

internal/api/api_default.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+func ConfigsGet(w http.ResponseWriter, r *http.Request) {
+	fmt.Printf("Starting server at port 8081\n")
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+
+	w.WriteHeader(http.StatusOK)
+
+	configValues := []Config{
```
converge, line 26: we can do it later: the function is a little big, we can improve it together on a later point.

internal/api/swagger.yaml
```diff
@@ -0,0 +1,87 @@
...
+    description: Find out more
+    url: https://kubefirst.com
+- name: configs
+  description: Viper configs from .kubefirst
+paths:
+  /configs:
+    summary: Gets a set of viper configs
+    get:
+      summary: ""
+      description: ""
```
converge, line 30: it would be nice to have some details for the API user/reader

internal/api/swagger.yaml
```diff
@@ -0,0 +1,87 @@
...
+                Config response:
+                  value:
+                  - key: configKey
+                    value: configValue
+            application/xml:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Config'
+        "404":
```
converge, line 52: on a later point: we could include examples of the http responses.

cmd/api.go
```diff
@@ -0,0 +1,27 @@
...
+var api = &cobra.Command{
+	Use:   "api",
+	Short: "starts API server",
+	Long:  "starts intermal API server",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		log.Printf("Server started")
+
+		router := sw.NewRouter()
+	
+		log.Fatal(http.ListenAndServe(":8081", router))
```
converge, line 19: panic is usually better than fatal, fatal will do a "force quit" `exit(1)` , and panic will call the `defers` / close open connection, files, etc before exiting.

pagottoo (dismissed):

internal/api/api_default.go
CristhianF7: Yeah. I thought the same. I'll definitely refactor this in a following PR

cmd/api.go
CristhianF7: Good to know this. Great catch 🚀

converge (approved):

6za (approved):

Issue #339, opened 2022-09-08
1.9.x - Wire Installer UI


6za:
# Overview

Add to `cluster create` the calls to start API/UI for handoff screen, add it under a test-flag.

6za:
Already addressed on #348

PR #340, opened 2022-09-08
#339 start - Wire UI to Create under flag

6za:
Explore how to Wire UI to create


CristhianF7 (dismissed):

CristhianF7 (dismissed):

6za:
Added info on handoff screen: 
image
Console is working: 
image

cmd/create.go
```diff
@@ -60,10 +65,30 @@ var createCmd = &cobra.Command{
...
 
 		sendCompleteInstallTelemetry(globalFlags.DryRun, globalFlags.UseTelemetry)
-		reports.HandoffScreen(globalFlags.DryRun, globalFlags.SilentMode)
-		time.Sleep(time.Millisecond * 2000)
 		log.Println("Kubefirst installation finished successfully")
 		informUser("Kubefirst installation finished successfully", globalFlags.SilentMode)
+		log.Println("Skiping the presentation of console and api for the handoff screen")
+		if createFlags.EnableConsole {
+			log.Println("Starting the presentation of console and api for the handoff screen")
+			go func() {
```
converge, line 73: the goroutines are missing the join point

6za: What I am missing here?

On this use case, I am using the handoof screen to keep this routines alive. Both create a server that never returns unless there is a panic. 

Once we we close the handoff screen both as destroyed as desired. 

Do I really need the join point? What that helps on that scenario? I am may be missing something, but I am not clear on what.

6za: I disable this until we reach agreement on the feature, although it works as design. I need to merge this PR to get the other fix it has to unblock github install tests.

pagottoo (dismissed):
lgtm

pagottoo (approved):

Issue #337, opened 2022-09-08
1.9.x - Terraform update vault for secret SSH_PRIVATE_KEY


6za:
# Overview

https://github.com/kubefirst/gitops-template/pull/99#discussion_r965940376

pagottoo:
closed by: https://github.com/kubefirst/gitops-template/pull/100

6za:
Secret is not passing to vault side.

6za:
Tested on #340.

PR #348, opened 2022-09-09
Re-enable #337

6za:
Only #337 as it was removed from #340

By design this PR enabled Installer browser UI via a Flag(opt-in). 
It uses the handoff screen to keep the UI servers on, once the user leaves the handoff screen the is closed, the API and UI servers are shotdown. 

On silent mode, it is not tested, the idea was to just start and close the servers. But maybe we may have different ideas for it. 

How to use it: 
```bash 
./kubefirst cluster create --enable-console
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -68,26 +68,26 @@ var createCmd = &cobra.Command{
 		log.Println("Kubefirst installation finished successfully")
 		informUser("Kubefirst installation finished successfully", globalFlags.SilentMode)
 		log.Println(createFlags.EnableConsole)
-		/*
-			if createFlags.EnableConsole {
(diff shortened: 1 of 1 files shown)


6za:
This feature is a way to use in experimental mode to test the UI on github/gitlab installs until we work to have console working as designed.

6za:
@CristhianF7 I had to disable this on the previous PR, due to lack of consensus. You can use this branch to test create with UI.

6za:
As this still experimental, the idea is to have a [feature flag](https://www.atlassian.com/continuous-delivery/principles/feature-flags) to allows developers to explore/sandbox the UI, before we commit with the design.

converge (approved):
approved with notes https://github.com/kubefirst/kubefirst/issues/349

Issue #354, opened 2022-09-09
1.9.x - Update destroy to skip destroy argo calls


6za:
# Overview 

#353  Adds a situation we can't skip the disable auto-synch, if that is still in `main`, the goal of this issue to PR a set of flags to skip this action. To prevent issues on destroy.

6za:
PR created

PR #356, opened 2022-09-09
#354 Add flag wrapping

6za:
- #354 to prevent issues when argo is not deployed or already destroyed. 

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -3,6 +3,11 @@ package cmd
 import (
 	"bytes"
 	"fmt"
+	"log"
+	"os/exec"
+	"syscall"
+	"time"
+
 	"github.com/kubefirst/kubefirst/configs"
 	"github.com/kubefirst/kubefirst/internal/argocd"
 	"github.com/kubefirst/kubefirst/internal/flagset"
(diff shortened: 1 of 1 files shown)


cmd/destroy.go
```diff
@@ -131,25 +132,28 @@ if the registry has already been deleted.`,
 
 		log.Println("gitlab terraform destruction complete")
 
-		log.Println("disabling ArgoCD auto sync")
-		argoCDUsername := viper.GetString("argocd.admin.username")
-		argoCDPassword := viper.GetString("argocd.admin.password")
+		//This should wrapped into a function, maybe to move to: k8s.DeleteRegistryApplication
```
converge, line 135: side note; it would be nice to stick in the cmd/driver, move functions into smaller functions is nice, but from an architecture perspective, when we introduce methods, it will become "macarrônico".

pagottoo (approved):
lgtm

cmd/destroy.go
6za: I am not getting why functions are bad... this should be part of delete register, as this  happens only `k8s.DeleteRegistryApplication` is executed. 

I am don't share what arch choice you are referring, anyway.. original code is breaking destroy. We are proceeding with merge to unblock some work. ;)

6za: I just want want stuff that run together, inside the same "if" to be near, so the behavior is consistent.

PR #357, opened 2022-09-09
feat: add lets encrypt validation during installation

### LINKED ISSUE #193: AwaitHostNTimes needs to use https client
AwaitHostNTimes `(internal/gitlab)` uses `http.Get()` that accepts HTTP or HTTPS when requesting HTTP. We should only accept valid HTTPS responses adding `Go http default client: http.DefaultClient` like we do on `getArgoCDToken` function.

converge:
- add lets encrypt validation during installation

close #193 

### DIFF
--- a/cmd/createGitlab.go
+++ b/cmd/createGitlab.go
@@ -214,6 +214,13 @@ var createGitlabCmd = &cobra.Command{
 
 			informUser("Waiting for Gitlab dns to propagate before continuing", globalFlags.SilentMode)
 			gitlab.AwaitHost("gitlab", globalFlags.DryRun)
+
+			domain := viper.GetString("aws.hostedzonename")
+			isValidCertificate, err := pkg.AwaitValidLetsEncryptCertificateNTimes(domain, globalFlags.DryRun, 60)
+			if err != nil || !isValidCertificate {
+				log.Println(err)
+			}
+
 			informUser("Pushing gitops repo to origin gitlab", globalFlags.SilentMode)
 			// refactor: sounds like a new functions, should PushGitOpsToGitLab be renamed/update signature?
 			viper.Set("gitlab.oidc-created", true)
--- a/cmd/createGitlab.go
+++ b/cmd/createGitlab.go
@@ -214,13 +214,6 @@ var createGitlabCmd = &cobra.Command{
 
 			informUser("Waiting for Gitlab dns to propagate before continuing", globalFlags.SilentMode)
 			gitlab.AwaitHost("gitlab", globalFlags.DryRun)
-
-			domain := viper.GetString("aws.hostedzonename")
-			isValidCertificate, err := pkg.AwaitValidLetsEncryptCertificateNTimes(domain, globalFlags.DryRun, 60)
-			if err != nil || !isValidCertificate {
-				log.Println(err)
-			}
-
 			informUser("Pushing gitops repo to origin gitlab", globalFlags.SilentMode)
 			// refactor: sounds like a new functions, should PushGitOpsToGitLab be renamed/update signature?
 			viper.Set("gitlab.oidc-created", true)
--- a/pkg/helpers.go
+++ b/pkg/helpers.go
@@ -1,7 +1,6 @@
 package pkg
 
 import (
-	"crypto/tls"
 	"errors"
 	"fmt"
 	"github.com/kubefirst/kubefirst/configs"
@@ -13,7 +12,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"time"
 )
 
 type RegistryAddon struct {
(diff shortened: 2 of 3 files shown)


cmd/createGitlab.go
```diff
@@ -214,6 +214,13 @@ var createGitlabCmd = &cobra.Command{
 
 			informUser("Waiting for Gitlab dns to propagate before continuing", globalFlags.SilentMode)
 			gitlab.AwaitHost("gitlab", globalFlags.DryRun)
+
+			domain := viper.GetString("aws.hostedzonename")
+			isValidCertificate, err := pkg.AwaitValidLetsEncryptCertificateNTimes(domain, globalFlags.DryRun, 60)
+			if err != nil || !isValidCertificate {
+				log.Println(err)
```
6za, line 221: just valid certificate would be enough. We don't need the full `AwaitValidLetsEncryptCertificateNTimes`

6za: The issue is for valid certificates in general.

converge: do you mean, one single request to see if the certificate is invalid?

pkg/helpers.go
```diff
@@ -261,3 +262,65 @@ func CreateFullPath(p string) (*os.File, error) {
...
+func AwaitValidLetsEncryptCertificateNTimes(domain string, dryRun bool, maxTimes int) (bool, error) {
+	log.Println("AwaitValidTLSCertificateNTimes called")
+	if dryRun {
+		log.Printf("[#99] Dry-run mode, AwaitValidTLSCertificateNTimes skipped.")
+		return true, nil
+	}
+	for i := 0; i < maxTimes; i++ {
+		validCertificate, err := IsLetsEncryptCertificateDomain(domain)
+		if err != nil {
+			return false, err
```
6za, line 276: the test is not if the domain is right, if if is valid certificate. `Curl -k` vs `plain curl`

6za: I think you went a bit too deep on this, the idea is to fail when the certificate is not valid. Broken Lock. 


As we allow insecure and it seems to be global.

converge: I didn't get it, could you re-phase..?

6za: From curl documentation, it would better than me explaining. 

```
       -k, --insecure
              (TLS) By default, every SSL connection curl makes is verified to be secure. This option allows curl to proceed and operate even
              for server connections otherwise considered insecure.

              The server connection is verified by making sure the server's certificate contains the right name and verifies successfully using
              the cert store.

              See this online resource for further details:
               https://curl.se/docs/sslcerts.html

              WARNING: this makes the transfer insecure.

              Example:
               curl --insecure https://example.com

              See also --proxy-insecure and --cacert. 
```

6za: This is a broken lock:
image

6za: This is a valid lock: 
image


Can you see the difference on the browser?  there is a warning on the address bar..

6za: Our current client is working with `--insecure` enabled, meaning we don't validate the cert against a valid cert-chain of trust. It is nothing to do with `let's encrypt` we can mock and "fake be it" and still have an invalid certificate. 


It is a bultin behavior of any http lib, that supports https to handle the cert check.

6za: The code in this PR diverges to this direction, creating a custom cert validation rule with limited scope, and not what was requested. I send an invite to a time in Monday to review together and understand better the intent of this PR.

cmd/createGitlab.go
6za: I refer to the long name of the function, details on the next comment, the issue is not about let's encrypt it is about certs that are trusted.

cmd/createGitlab.go
```diff
@@ -214,6 +214,13 @@ var createGitlabCmd = &cobra.Command{
...
+			}
+
 			informUser("Pushing gitops repo to origin gitlab", globalFlags.SilentMode)
 			// refactor: sounds like a new functions, should PushGitOpsToGitLab be renamed/update signature?
```
6za, line 225: Let's avoid to refactor the the places we call this method. It is a new function, for selected places. Let's focus on just adding the new function on this PR. I newer PR after we stabilize main will use this function on selected places. 

After some testing of behavior impact.

converge:
@6za is there something else you want other than my initial ticket description? https://github.com/kubefirst/kubefirst/issues/193#issue-1329044637

6za:
I will test this branching on my next available window. I want to see if changes the behavior on the enf of the install when we wait for argo to be ready.

6za (changes requested):
Blocking merge, as impact installer flow. Let's put on hold this effort. Too much effort spent on this PR, and it takes some resources/time to test. 

Next run need to be done with new certs.

6za:
We may want to close this PR without merge, we will explore a different path of adding more functions on this library: https://github.com/kubefirst/kubefirst/blob/main/internal/httpCommon/client.go#L11

To make more generic HTTP Call checkers and based on the use case, to trust or not on  `insecure pipes`. 

The original intent of the issue #193  was to create a more generic constructor for HTTP endpoints, to allow opt-in of insecure mode when desired. Preserving the original behaviors. 

This will be tackled on future attempts of #193

converge:
> We may want to close this PR without merge, we will explore a different path of adding more functions on this library: https://github.com/kubefirst/kubefirst/blob/main/internal/httpCommon/client.go#L11

this function creates a new transport and HTTP client every time it's called. we should have a single http client for the whole driver/command. https://godoc.org/net/http#Transport

converge:
> _The function mentioned before is just meant to create the client, not necessarily to be used on every call. Anyway, not that is not focus of this PR and this PR not the place to discuss future solutions. This code was late for when it was needed._
> 
> **As installer is working as expected for 1.9.0 at the moment, my personal opinion this may introduce bugs that we don't needed close to the release. So, I see as a waste of effort to try to get this merged, that could be better spent trying to improve the rest of the issues open at the moment.**
> 
> **If you desire to do so, it is up to you, ensure `gitlab` and `github` flows are tested to create/destroy before merging and after merging. So, we don't impact `main` with new bugs.**
> 
> I agree, that is preferred to avoid create new instances of objects when not needed. Although that was not the main goal of #193. Its main goal was simple mechanism to prevent `insecure pipes` to be trusted. Several iterations of this PR were done to try to achieve that without success.
> 
> As it seems, maybe the issue/ticket was not clear, or there was a lack of communication to confirm the objective of the ticket. Either way, the installer today is working well as is. In the future, I have been trying to make tickets more clear to avoid waste of cycles on this type of back/forth.

my comment was related to the direction of the previous comment regarding re-creating new http clients on every http request. I agree with the rest, communication was funky at this ticket, we ended up doing the initial ticket description.

PR #407, opened 2022-09-16
Feat ci release tag

### LINKED ISSUE #252: Create CI/CD automations for release tagging
## Background

Today in order to have a release cut we need:
- We need to trigger release creation(on kubefirst/kubefirst)
-  Update `config.KubefirstVersion` : https://github.com/kubefirst/kubefirst/blob/main/configs/config.go#L80
- We need to tag all template repos with the same version of release: 
  - https://github.com/kubefirst/gitops-template
  - https://github.com/kubefirst/metaphor-template
  - https://github.com/kubefirst/metaphor-go-template
  - https://github.com/kubefirst/metaphor-frontend-template

- We need tu publish a docker hub image for: kubefirst/kubefirst

# Goal of this ticket:

To create a single action that updates `cli` version by generating a file and from there tagging all repositories freezing release assets.

pagottoo:
Close #252 

With this PR we introduce the mechanism to: 

- build the binaries with the version according to the release name;
- build/push docker image to registry;
- update the docker hub repo;
- tag all template repositories with the same version of the release.


pagottoo:
Before merging, we need to confirm if the project has the tokens to perform operations against the GH API and docker hub.

configs/config.go
```diff
@@ -14,6 +14,9 @@ This is an initial implementation of Config. Please keep in mind we're still wor
 environment variables and general config data.
 */
 
+// to use with parans ldlfags
+var k1Version string
+
 // Config host application configuration
 // todo: some of these values can be moved to the .env
```
6za, line 21: **How we run for development?**

Do we just export this on our terminal before runs:
```
export LDFLAGS="-X github.com/kubefirst/kubefirst/configs.k1Version=1.8.6"
```

6za (approved):
Please, communicate the new step to run with the team in the action channel. @marinhojess may need to update some docs after this PR is merged. 

Great work!

converge:
+1 to add some docs.

fyi; the `version` command is returning empty for the version

configs/config.go
```diff
@@ -14,6 +14,9 @@ This is an initial implementation of Config. Please keep in mind we're still wor
 environment variables and general config data.
 */
 
+// to use with parans ldlfags
+var k1Version string
```
converge, line 18: is this var receiving the value from a environment variable?

6za: it seems to be this one: `export LDFLAGS="-X github.com/kubefirst/kubefirst/configs.k1Version=1.8.6"`

6za: A building flag mechanism.

converge:
> > +1 to add some docs.
> > fyi; the `version` command is returning empty for the version
> 
> Did you exported the LDFLAGS when running info?

dont know where the version should come from, but the version command should return the version.

converge:
in order to not break the `go run . version` command, we could:
- `const K1Version = "development"`
and update the `version` command to read the version from `K1Version`. 

note; `K1Version` must be uppercase to be exported, and `version` command be able to read it from another package.

it will give us a "developer" version, and a built version that will use the `LDFLAG`

pagottoo:
Hey team, I tested with const and LDFLAGs cannot overwrite the value, so I continue using var with value `main` because e don't have the `development` branch in our repos, so when a contributor or kubefirst team is developing we can use the `main` as branch base.

converge (approved):
awesome!

6za (approved):

Issue #386, opened 2022-09-12
github install - chartmusuem not available yet.


6za:
# Overview

As is, metaphors application once pushed will execute the build job, and if that happens too early chartmuseum is not ready yet. 

```bash 
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-908752723: version: 0.1.0-rc.b9f797532946451b3c226fbce2d7b324f
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832: Error: looks like "https://chartmuseum.6zray.kubefast.com/" is not a valid chart repository or cannot be reached: Get "https://chartmuseum.6zray.kubefast.com/index.yaml": dial tcp: lookup chartmuseum.6zray.kubefast.com on 172.20.0.10:53: no such host
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832: Error: couldn't load repositories file (/root/.config/helm/repositories.yaml): open /root/.config/helm/repositories.yaml: no such file or directory
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832: Usage:
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832:   helm push [flags]
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832: 
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832: Flags:
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832:       --access-token string             Send token in Authorization header [$HELM_REPO_ACCESS_TOKEN]
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832:       --auth-header string              Alternative header to use for token auth [$HELM_REPO_AUTH_HEADER]
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832:       --ca-file string                  Verify certificates of HTTPS-enabled servers using this CA bundle [$HELM_REPO_CA_FILE]
metaphor-publish-b9f797532946451b3c226fbce2d7b324fz2ftz-1555305832:       --cert-file string                Identify HTTPS client using this SSL certificate file [$HELM_REPO_CERT_FILE]
[...]

PR #410, opened 2022-09-16
Check status of install

6za:
Solves:
- #386 

Creates a command to check if the tolling k1 need to ensure metaphors will work is already synched.

Success Logs sample:
```bash
LOG: 2022/09/16 18:45:46.592113 /home/developer/app/cmd/k1Ready.go:27: argocdAppStatus called
LOG: 2022/09/16 18:45:46.592570 /home/developer/app/internal/flagset/common.go:58: Enviroment Variable(dry-run) not set
LOG: 2022/09/16 18:45:46.593009 /home/developer/app/internal/flagset/common.go:58: Enviroment Variable(use-telemetry) not set
LOG: 2022/09/16 18:45:46.593408 /home/developer/app/internal/flagset/common.go:58: Enviroment Variable(silent) not set
LOG: 2022/09/16 18:45:46.594490 /home/developer/app/cmd/k1Ready.go:48: argo forwarded called
LOG: 2022/09/16 18:45:49.440722 /home/developer/app/internal/argocd/argocd.go:454: App status: Synced
LOG: 2022/09/16 18:45:49.441154 /home/developer/app/cmd/k1Ready.go:61: App registry is synched: true
LOG: 2022/09/16 18:45:50.616328 /home/developer/app/internal/argocd/argocd.go:454: App status: Synced
LOG: 2022/09/16 18:45:50.633384 /home/developer/app/cmd/k1Ready.go:61: App argocd is synched: true
LOG: 2022/09/16 18:45:51.588172 /home/developer/app/internal/argocd/argocd.go:454: App status: Synced
LOG: 2022/09/16 18:45:51.588843 /home/developer/app/cmd/k1Ready.go:61: App atlantis is synched: true
LOG: 2022/09/16 18:45:52.722477 /home/developer/app/internal/argocd/argocd.go:454: App status: Synced
LOG: 2022/09/16 18:45:52.722921 /home/developer/app/cmd/k1Ready.go:61: App cert-manager is synched: true
LOG: 2022/09/16 18:45:53.743629 /home/developer/app/internal/argocd/argocd.go:454: App status: Synced
LOG: 2022/09/16 18:45:53.744314 /home/developer/app/cmd/k1Ready.go:61: App chartmuseum is synched: true
LOG: 2022/09/16 18:45:54.755091 /home/developer/app/internal/argocd/argocd.go:454: App status: Synced
LOG: 2022/09/16 18:45:54.755570 /home/developer/app/cmd/k1Ready.go:61: App chartmuseum-components is synched: true

```

### DIFF
--- a/internal/argocd/argocdStructs.go
+++ b/internal/argocd/argocdStructs.go
@@ -1,0 +1,255 @@
+package argocd
+
+import "time"
+
+type ArgoCDConfig struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+
+type SyncResponse struct {
+	Status struct {
+		Sync struct {
+			Status string `json:"status"`
+		} `json:"sync"`
+	}
+}
+
+// Application is required with full specification since ArgoCD needs a PUT to update the syncPolicy, and there is no
+// PATCH available
+type Application struct {
+	Metadata struct {
+		Name              string    `json:"name"`
+		Namespace         string    `json:"namespace"`
+		Uid               string    `json:"uid"`
+		ResourceVersion   string    `json:"resourceVersion"`
+		Generation        int       `json:"generation"`
+		CreationTimestamp time.Time `json:"creationTimestamp"`
+		ManagedFields     []struct {
+			Manager    string    `json:"manager"`
+			Operation  string    `json:"operation"`
+			ApiVersion string    `json:"apiVersion"`
+			Time       time.Time `json:"time"`
+			FieldsType string    `json:"fieldsType"`
+			FieldsV1   struct {
+				FSpec struct {
+					Field1 struct {
+					} `json:"."`
+					FDestination struct {
+						Field1 struct {
+						} `json:"."`
+						FNamespace struct {
+						} `json:"f:namespace"`
+						FServer struct {
+						} `json:"f:server"`
+					} `json:"f:destination"`
+					FProject struct {
+					} `json:"f:project"`
+					FSource struct {
+						Field1 struct {
+						} `json:"."`
+						FPath struct {
+						} `json:"f:path"`
+						FRepoURL struct {
+						} `json:"f:repoURL"`
+					} `json:"f:source"`
+					FSyncPolicy struct {
+					} `json:"f:syncPolicy"`
+				} `json:"f:spec,omitempty"`
+				FStatus struct {
+					Field1 struct {
+					} `json:".,omitempty"`
+					FHealth struct {
+						FStatus struct {
+						} `json:"f:status,omitempty"`
+					} `json:"f:health"`
+					FSummary struct {
+						FImages struct {
+						} `json:"f:images,omitempty"`
+					} `json:"f:summary"`
+					FSync struct {
+						Field1 struct {
+						} `json:".,omitempty"`
+						FComparedTo struct {
+							Field1 struct {
+							} `json:".,omitempty"`
+							FDestination struct {
+								FNamespace struct {
+								} `json:"f:namespace,omitempty"`
+								FServer struct {
+								} `json:"f:server,omitempty"`
+							} `json:"f:destination"`
+							FSource struct {
+								FPath struct {
+								} `json:"f:path,omitempty"`
+								FRepoURL struct {
+								} `json:"f:repoURL,omitempty"`
+							} `json:"f:source"`
+						} `json:"f:comparedTo"`
+						FRevision struct {
+						} `json:"f:revision,omitempty"`
+						FStatus struct {
+						} `json:"f:status,omitempty"`
+					} `json:"f:sync"`
+					FHistory struct {
+					} `json:"f:history,omitempty"`
+					FOperationState struct {
+						Field1 struct {
+						} `json:"."`
+						FFinishedAt struct {
+						} `json:"f:finishedAt"`
+						FMessage struct {
+						} `json:"f:message"`
+						FOperation struct {
+							Field1 struct {
+							} `json:"."`
+							FInitiatedBy struct {
+								Field1 struct {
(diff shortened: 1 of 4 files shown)


6za:
By design command will fail in case, k1 tolloing is not ready.

internal/argocd/argocdStructs.go
```diff
@@ -0,0 +1,255 @@
+package argocd
```
converge, line 1: idiomatic Go have types closer to their implementation, it's a bit bad to have to open another file to check the struct definition, I would like the idiomatic way

6za: Reading side by side is easier.. is go fan of 100s of line files?
Let me check k8s packages, I saw something different on some projects I checked.

6za: Do you have any reference for this style, searching for it.

converge: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cronjob/cronjob_controllerv2.go

6za: nothing here: https://github.com/golovers/effective-go

6za: nothing that calls out here: https://go.dev/doc/effective_go#package-names

converge: > Reading side by side is easier.. is go fan of 100s of line files?
better a big file than trying to find what you need in different files. following that idea we soon will have
```
package-name/functions.go
package-name/types.go
package-name/methods.go
```

or
```
package-name/my-func.go
package-name/my-other-func.go
```

converge: maybe the day Kubefirst has a big chunk of types that are third party packages we can have it all on a single `types.go`, but it doesn't make much sense to do it for small packages. the argo package has one single file

6za: https://github.com/go-git/go-git/tree/master/internal/revision
For example: 
image

The idea was to isolate the code that is generated by yaml to struct from human code. 

On ideal design, as you said they would come from importing 3rd part pkg or isolating common data types on type packages k8s does. 

To avoid bloated code. 

As this is a candidate to PR conflict nightmare as size of file mix actions. I would agree to separate into smaller files by them, like application level action and argoCD global level actions, based on a semantic separation. 

But mix human/ tool generate code may bloat the code in a unnecessary way.

6za: Another sample: https://github.com/kubernetes/client-go/tree/master/applyconfigurations/core/v1

image

6za: Where to avoid a huge code base, and PR nightmare from multiple source an strategy of breaking smaller concerns were applied

6za: This struct are meant to expected to be updated as argo update its apis, the logic may persist for longer if the struct has only incremental changes.

converge: where does these examples have a file with structs only?

6za: Pkg that were broken into files to make easier to handle code that is growing.. I agree on having a definition of how to break it, but to farm huge files with huge structs is not a good practice.

converge: in this specific case (argo struct), I agree, the struct is huge, but there is a reason for that.

Argo API doesnt provide an endpoint to PATCH the application, for that we need to follow what Argo UI does that is firstly `GET` -> store it in the struct, and `PUT`.

I also a fan to have smaller functions and isolate, but there are better strategies for that.

6za:
The code was refactored a bit the naming to match the names of structs to match the Swagger spec from argo. 
https://github.com/argoproj/argo-cd/blob/v2.4.6/assets/swagger.json

Although, after some tests, of raw translation: 
```bash 
docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli generate \
    -i /local/api.json \
    --model-package argocdModel \
    -l go \
    -o /local/out/go
```

The raw spec from swagger seems to not match when translated to go the expectations to proper decoding. 

So, following a code style similar to k8s APIs specs:
https://github.com/kubernetes/client-go/tree/master/applyconfigurations/core/v1

It may be better to not mix "external types". with our client libraries. 

It was created a file for each struct to create "common" way to separate concerns, in the future I think this kind of code is ephemeral as they should be imported from tools client **libraries packages** or **automatic generated** with tools similar the one above. 


Most of the "package" discussion above, for me seems pointless, as it seems to be lack of shared set of tastes and style code wise: https://github.com/kubefirst/kubefirst/pull/410#discussion_r973313710


As no references were provided of a good project organization, my goal and intent is to provide a way it easy to update types, easy to  avoid PR conflicts and focus on smaller source code files. 

**Principles in play:**
- **Longer source files are more prone to create merge conflicts and longer time for PR merge**
- **Holding huge "translated" code parts into code generate by humans, makes harder to automate updates of this files via automated means.** 
- **Isolate common code to make easier future refactories** 

_To just say it is an idiomatic-way without sources,_ I think is not a nice way to help others to converge to a common way of work and produce similar code, as every developer has a different code journey. Taste is tricky thing to share.

internal/argocd/argocdStructs.go
6za: https://github.com/kubefirst/kubefirst/pull/410#issuecomment-1249977388

6za:
And I would add to that due to my long experience watching and using [MVC](https://www.codecademy.com/article/mvc) Model and ControlLer(logic) it is hard to seem the `structs` that for api works as models to be mixed with controller, that usually makes refactoring very complicate.

PR #419, opened 2022-09-19
Explore #386

6za:
Attempt to get closer to fix:
- #386  

This help to add extra checks on CM status, but it seems to doesn't solve #386 yet. Some changes need to happen in template side. 


6za:
This Pr also introduce a new pkg to handle ChartMuseum checking library, that may grow to handle CM utilities.

pagottoo (approved):
lgtm

PR #433, opened 2022-09-21
Fix atlantis GitHub webhook

### LINKED ISSUE #398: 1.9.x - ensure atlantis is wired properly on github installs
# Overview 

Based on tests, it seems atlantis is not properly wired into github installs. 

The goal of this task is to ensure feature parity between gitlab/github installs, meaning that atlantis works as in gitlab inside a github installation. 

Places of interest top review:
https://github.com/kubefirst/gitops-template/blob/main/terraform/vault/bootstrap/secrets.tf#L4
> secrets loaded into atlantis 

https://www.runatlantis.io/docs/server-configuration.html
> Atlantis configs

https://github.com/kubefirst/gitops-template/blob/main/components/atlantis/application.yaml#L22
> Atlantis helm values, initial setup properties.

pagottoo:
Closes #398 

With this PR we create a webhook configuration in the gitops repo, set a passcode, and inform the terraform vault modules to be used on Atlantis configuration.

### DIFF
--- a/pkg/helpers.go
+++ b/pkg/helpers.go
@@ -5,9 +5,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/kubefirst/kubefirst/configs"
 	"github.com/spf13/viper"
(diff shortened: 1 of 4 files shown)


6za (approved):
lgtm

pkg/helpers.go
```diff
@@ -25,6 +27,8 @@ type RegistryAddon struct {
 	} `yaml:"metadata"`
 }
 
+var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
converge, line 30: is `letters` only used in `randSeq()` function?

maybe it's a good idea to include it in the function scope like:
```go
func randSeq( ... ) ... {
  var letter = []rune( ... )
  runeSlice := make ...
  ...
}
```

in this way we avoid having global scope variable.

converge (dismissed):
added a comment, feel free to use it or not

6za (approved):

Issue #416, opened 2022-09-19
Destroy - Disable auto-synch doesn't work as expected for github install


6za:
```bash 
LOG: 2022/09/19 18:45:18.672546 /home/developer/app/internal/k8s/kubernetes.go:193: /home/developer/.k1/tools/kubectl   --kubeconfig   /home/developer/.k1/gitops/terraform/base/kubeconfig   -n   argocd   port-forward   svc/argocd-server   8080:80
LOG: 2022/09/19 18:45:18.672855 /home/developer/app/cmd/createUtils.go:303: Open argocd port-forward
LOG: 2022/09/19 18:45:18.673209 /home/developer/app/cmd/destroy.go:125: disabling ArgoCD auto sync
LOG: 2022/09/19 18:45:22.747817 /home/developer/app/cmd/destroy.go:145: Error stopping synch of Registry: unable to update ArgoCD application, http response code is 404
LOG: 2022/09/19 18:45:22.748788 /home/developer/app/cmd/destroy.go:118: Closed argocd port forward
LOG: 2022/09/19 18:45:22.749711 /home/developer/app/cmd/destroy.go:103: Closed argo port forward

```

6za:
After disabled, the auto-synch is re-enabled. 

And current code on destroy is returning `404` on the PUT operation.

6za:
issue also happens on `gitlab` installs: 
```
developer@a986497c080c:~/app$ ./kubefirst destroy
2022/09/21 15:59:33 kubefirst started
Logging at: /home/developer/app/logs/log_1663775973.log
LOG: 2022/09/21 15:59:33.517613 /home/developer/app/pkg/helpers.go:251: Using config file: /home/developer/.kubefirst
Init actions: 2 expected tasks ...

Open Ports                    ... done! [3 in 5ms]
- Open gitlab port-forward
- Open argocd port-forward
- Open vault port-forward
- Destroying Gitlab
Destroy Cloud                 ... 50.0% [##############..............] [2 in 44.197027s]
[##################################################..................] [44s]
Error: unable to update ArgoCD application, http response code is 404
Usage:
  kubefirst destroy [flags]

```

6za:
## logs

```bash 
LOG: 2022/09/21 16:00:14.256495 /home/developer/app/cmd/destroy.go:133: gitlab terraform destruction complete
[...]

PR #435, opened 2022-09-21
chore: remove auto sync call due to argocd bug

converge:
I tried different approaches to solve it, but there is an ArgoCD bug[1] that always reconcile to the source desired stated.

approaches:
1. HTTP request the same way ArgoCD UI does
2. `kubectl -n <namespace> patch --type='merge' application <application-name> -p "{\"spec\":{\"syncPolicy\":null}}`

outcome: in both cases, it does update the state to auto sync disabled as we expect, but ArgoCD reconciliation auto enable auto sync it again.

[1] https://github.com/argoproj/argo-cd/issues/3895


6za:
Maybe we can close this PR, `destroy` was updated to solve issue produced on #416 - Further research will be done for alternative ways to destroy support removal of registry on the template side.

converge:
disable auto sync is dead code since the ArgoCD bug described above, it does disable auto-sync, but ArgoCD reconciliation re-enable it again. this will be in the git history and we can roll it back when ArgoCD fix the bug.

6za (approved):
Approving, please ensure this is tested and no bugs are added to main, so we keep ready for 1.9.0. 

I still don't understand how the issue was introduced in the first place, as it seems was not tested before, as the solution was not operational and failed on both type of install on destroy. 

We need to ensure **gitlab** and **github** flows are not impacted. Please, perform full install/destroy once this code is merged.

Issue #434, opened 2022-09-21
1.9.x - gitlab install argo cd missing for helm install


6za:
# Erro

```bash
panic: error: could not helm install argocd command exit status 1

goroutine 1 [running]:
log.Panicf({0x1c3ade3?, 0x1e?}, {0xc0000f3728?, 0xe?, 0xe?})
	/usr/local/go/src/log/log.go:392 +0x67
github.com/kubefirst/kubefirst/internal/helm.InstallArgocd(0x0)
	/home/developer/app/internal/helm/helm.go:32 +0x4a5
github.com/kubefirst/kubefirst/cmd.glob..func12(0xc000105860?, {0xc000255ab0, 0x0, 0x1})
	/home/developer/app/cmd/createGitlab.go:105 +0x6af
github.com/kubefirst/kubefirst/cmd.glob..func10(0x2b9fde0, {0xc000255ab0?, 0x0, 0x1})
	/home/developer/app/cmd/create.go:50 +0x1d9
github.com/spf13/cobra.(*Command).execute(0x2b9fde0, {0xc000255a90, 0x1, 0x1})
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0x2ba0060)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
github.com/kubefirst/kubefirst/cmd.Execute()
	/home/developer/app/cmd/root.go:26 +0x25
main.main()
```

## Logs

```bash 
LOG: 2022/09/21 16:41:13.815420 /home/developer/app/pkg/shell.go:24: ERR: 
LOG: 2022/09/21 16:41:14.204691 /home/developer/app/pkg/shell.go:22: Command Execution: /home/developer/.k1/tools/helm
LOG: 2022/09/21 16:41:14.205089 /home/developer/app/pkg/shell.go:23: OUT: Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "argo" chart repository
Update Complete. ⎈Happy Helming!⎈

LOG: 2022/09/21 16:41:14.205468 /home/developer/app/pkg/shell.go:24: ERR: 
[...]

Issue #429, opened 2022-09-21
1.9.x - gitlab install still takes a while to have softserve ready


6za:
# Error

```bash 
Init actions: 4 expected tasks ...

Process Parameters            ... done! [1 in 0s]
- Creating K8S Cluster
Prepare Temporary Repo        ... 25.0% [#######.....................] [1 in 13m44.901012s]
- Created Softserve
- Waiting Softserve
Prepare Temporary Repo        ... done! [4 in 14m31.228s]
panic: we tried hard to setup SoftServe but there were something wrong, please check the logs

goroutine 1 [running]:
log.Panic({0xc0000eb888?, 0x1c0f10c?, 0xc0000eb888?})
	/usr/local/go/src/log/log.go:385 +0x65
github.com/kubefirst/kubefirst/internal/softserve.ConfigureSoftServeAndPush(0x0)
	/home/developer/app/internal/softserve/softserve.go:96 +0x316
github.com/kubefirst/kubefirst/cmd.glob..func12(0xc0001056c0?, {0xc000201bf0, 0x0, 0x1})
	/home/developer/app/cmd/createGitlab.go:99 +0x5ff
github.com/kubefirst/kubefirst/cmd.glob..func10(0x2b9fde0, {0xc000201bf0?, 0x0, 0x1})
	/home/developer/app/cmd/create.go:50 +0x1d9
github.com/spf13/cobra.(*Command).execute(0x2b9fde0, {0xc000201be0, 0x1, 0x1})
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0x2ba0060)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
github.com/kubefirst/kubefirst/cmd.Execute()
	/home/developer/app/cmd/root.go:26 +0x25
main.main()
	/home/developer/app/main.go:65 +0x41d
```

## Logs
```bash 
LOG: 2022/09/21 12:37:37.814911 /home/developer/app/cmd/createUtils.go:303: Softserve Update
LOG: 2022/09/21 12:37:37.816504 /home/developer/app/internal/softserve/softserve.go:62: Executing configureSoftserveAndPush
[...]

PR #437, opened 2022-09-21
 #429 #434

6za:
Addresses some bug found:
- #429 
- #434 
- Also reverted the restriction to APIs as we don't have secrets anymore on UI(Browser Console)
- Added retrial logic for steps that appears flaky on gitlabInstall

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -10,8 +10,10 @@ import (
 )
 
 // InstallArgocd - install argoCd in a cluster
+// it has a retry embeded logic to mitigate network issues when trying to install argoCD
 func InstallArgocd(dryRun bool) error {
 	config := configs.ReadConfig()
+	message := "error installing argo-cd: unexpected state"
 	if !viper.GetBool("create.argocd.helm") {
 		if dryRun {
 			log.Printf("[#99] Dry-run mode, helmInstallArgocd skipped.")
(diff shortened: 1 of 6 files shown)


internal/helm/helm.go
```diff
@@ -9,33 +9,47 @@ import (
...
 		// ! commenting out until a clean execution is necessary // create namespace
-		_, _, err := pkg.ExecShellReturnStrings(config.HelmClientPath, "--kubeconfig", config.KubeConfigPath, "repo", "add", "argo", "https://argoproj.github.io/argo-helm")
-		if err != nil {
-			log.Panicf("error: could not run helm repo add %s", err)
-		}
+		// Refers to: https://github.com/kubefirst/kubefirst/issues/434
+		totalAttempts := 5
+		for i := 0; i < totalAttempts; i++ {
+			log.Printf("Installing Argo-CD, attempt (%d of %d)", i+1, totalAttempts)
+			_, _, err := pkg.ExecShellReturnStrings(config.HelmClientPath, "--kubeconfig", config.KubeConfigPath, "repo", "add", "argo", "https://argoproj.github.io/argo-helm")
```
6za, line 25: this loop need to be moved to the callers.. or the logic need to be reviewed.

6za: The returns on error will keel the loop, that was not the intended behavior.

6za: logic updated, it should be fine now.

pagottoo (approved):
lgtm

Issue #422, opened 2022-09-20
ArgoCD - self-signed certificate conflicting with cert-manager certificate secret


pagottoo:

6za:
# Overview 

Based on discussion


@jarededwards 
> https://github.com/kubefirst/gitops-template/blob/main/components/argocd-gitlab/application.yaml#L31
[application.yaml](https://github.com/kubefirst/gitops-template/blob/main/components/argocd-gitlab/application.yaml)

> yes, thats exactly what we’re referring to. the cert used to live in argocd-secret or something to that effect but since we’re using a different secret argocd automatically generates a self signed certificate in argocd-secret and for some reason argocd has a bug that it switches between the two certificates (the correct one from cert manager in argocd-server-tls and the incorrect self signed in argocd-secret )

> it was cycling between the two in a browser on page refresh, it was odd


@pagottoo 
> huuum, :thinking_face:
> got it
> we could disable tls on argocd-server, and manage tls isolated on ingress...
or we could update the argocd-secret with the cert, and mark these fields on argo to avoid conflicts like argo warns about out of sync


@jarededwards 
> yea im not sure what the best way is - when we deleted the incorrect tls.key and tls.crt  in the argocd-secret  the problem went away , so it might be just as good to remove the self signed cert so we can leave the application as is


@pagottoo 
> yes, looks like a good approach


Reference: https://argo-cd.readthedocs.io/en/stable/operator-manual/tls/#tls-certificates-used-by-argocd-server

PR #474, opened 2022-09-28
Fix certs conflict argocd

pagottoo:
With this PR we close https://github.com/kubefirst/kubefirst/issues/422 using the k8s go SDK performing a patch operation removing self-signed cert from argo-secret.

### DIFF
--- a/internal/k8s/kubernetes.go
+++ b/internal/k8s/kubernetes.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


internal/k8s/kubernetes.go
```diff
@@ -290,3 +302,57 @@ func WaitForGitlab(dryRun bool, config *configs.Config) {
...
+
+	payload := []PatchJson{{
+		Op:   "remove",
+		Path: field,
+	}}
+
+	clientset, err := GetClientSet(false)
+	if err != nil {
+		log.Panicf("Error creating k8s clientset : %s", err)
+		return err
```
converge, line 336: the error wont be returned since log.Panic will `print`, call `panic`, and stop the execution.

pagottoo: Ok, I will change it to `printf` because if this process fails is not tragical and we can continue the process.

converge (approved):

converge (approved):

Issue #465, opened 2022-09-27
Handle 404 error while destroy buckets


marinhojess:
# What?

When executing the kubefirst clean command with the --destroy-buckets flag, the command may break if any buckets have already been deleted from the cloud.

```
Error: error deleting bucket/objects/version, the resources may have already been removed, please re-run without flag --destroy-buckets and check on console
```

If the bucket has really been deleted from the cloud, a log with error 404 can be viewed in the application log:

```
LOG: 2022/09/27 16:06:16.430216 /home/jessicamarinho/Documents/kubefirst/internal/aws/aws.go:574: Failed to list objects: operation error S3: ListObjectsV2, https response error StatusCode: 404, RequestID: xxx, HostID: xxx, NoSuchBucket: 
```

In this case the ideal would be to proceed with the clean process without breaking the command.

# How?

Develop a mechanism to handle the 404 error and proceed with the clean process.

PR #475, opened 2022-09-28
feat(clean): Handle error 404

marinhojess:

Making clean proceed with clean process even if the bucket has already been deleted from the cloud.


marinhojess:
This PR is addressed to issue #465

6za (approved):
lgtm

converge (approved):
AWS functions should be independent since different commands can use the different AWS services. Let's say destroy now also uses the destroy bucket function, it will contain clean logic.

but don't worry about that in this PR, our current codebase is not clear about that, and we can improve it on a later time together.

Issue #511, opened 2022-10-06
new switch --bot-password needs correction


johndietz:
the new --bot-password switch that was recently introduced doesn't have the intended influence on the initial admin user created on a github stack. it was introduced as part of the correction from #510 . the workaround is to use the root vault token to get the kubefirst-bot password at /users/kbot. that password does work with user kbot via sso as an admin throughout the platform.

echoboomer:
This code has since been deprecated and is no longer used.

PR #517, opened 2022-10-06
 #511 bot pass mandatory

6za:
- #511 

### DIFF
--- a/internal/flagset/installer.go
+++ b/internal/flagset/installer.go
@@ -1,6 +1,7 @@
 package flagset
 
 import (
+	"fmt"
 	"log"
 
 	"github.com/kubefirst/kubefirst/configs"
@@ -140,10 +141,9 @@ func ProcessInstallerGenericFlags(cmd *cobra.Command) (InstallerGenericFlags, er
 	log.Println("option.kubefirst.experimental", experimentalMode)
 	flags.ExperimentalMode = experimentalMode
 
-	// TODO: reintroduce the next 3 lines after #511 is closed
-	//if viper.GetBool("github.enabled") && flags.BotPassword == "" {
-	//	return InstallerGenericFlags{}, fmt.Errorf("must provide bot-password argument for github installations of kubefirst")
-	//}
+	if viper.GetBool("github.enabled") && flags.BotPassword == "" {
+		return InstallerGenericFlags{}, fmt.Errorf("must provide bot-password argument for github installations of kubefirst")
+	}
 
 	return experimentalModeTweaks(flags), nil
 }


internal/flagset/installer.go
```diff
@@ -140,10 +141,9 @@ func ProcessInstallerGenericFlags(cmd *cobra.Command) (InstallerGenericFlags, er
 	log.Println("option.kubefirst.experimental", experimentalMode)
 	flags.ExperimentalMode = experimentalMode
 
-	// TODO: reintroduce the next 3 lines after #511 is closed
-	//if viper.GetBool("github.enabled") && flags.BotPassword == "" {
-	//	return InstallerGenericFlags{}, fmt.Errorf("must provide bot-password argument for github installations of kubefirst")
-	//}
+	if viper.GetBool("github.enabled") && flags.BotPassword == "" {
+		return InstallerGenericFlags{}, fmt.Errorf("must provide bot-password argument for github installations of kubefirst")
```
converge, line 145: unit test will complain of Errorf, errors.New() will do the trick

6za: test just passed: 
image
Do we need to change? Is there any real impact?

6za: It probably complains when used Upper case later on the message.

converge: no real impact, but its nice to use the correct function for the correct job. 

in short, when there is a "f" in the function name in the `fmt` package, it usually means we can pass arguments to the string like ("%d", myInt).

https://pkg.go.dev/fmt#Errorf

pagottoo (approved):

internal/flagset/installer.go
converge: feel free to merge with this warning, I can fix it later

pagottoo (approved):
lgtm

6za:
Closing this PR as is to early to have this enabled.

internal/flagset/installer.go
6za: I added a discussion here if there are mode details on the choice: https://github.com/kubefirst/kubefirst/discussions/533

_I think language is tricky, I don't subscribe on the idea of it been "wrong" to be worth to be fixed. It seems to be at most a change with marginal performance gain._   

Anyway, if these is a new standard to be followed let's try to communicate the value of the change and if it is worth the refactory. Not all rules can be simply generalized, if so we need to context to follow the decision. 

As if you see only this line in the PR, it may be be "simple change", but as the rest of the code base also use the  "old way", so does this means we need to change all code we touch that has the older way? 

In theory, this PR just "uncommented" a part of the code.

Issue #406, opened 2022-09-15
Error to create cluster with assume role


marinhojess:
When using assume role with the Kubefirst installer running on Argo Workflows, it fails when trying to create the cluster.

![image](https://user-images.githubusercontent.com/19623668/190499991-865b3b7b-3a56-4961-b00b-7ca17a2a9898.png)

marinhojess:
We need to check if the AWS_PROFILE variable is conflicting with the ARN_ROLE authentication.

marinhojess:
The init command doesn't allow to be executed without the aws profile being informed, this is probably creating the authentication conflict.

![image](https://user-images.githubusercontent.com/19623668/190512477-a8f52acd-909d-4f19-9530-dfa25b9a0615.png)

6za:
Based on the last tests, we are having some challenge to confirm `assumeRole` works as expected for `init` operation. Although, it seems to work as expected on the `awswhoami` command.

marinhojess:
Added the following trusted relationship to the role that the CI wants to assume:

```
{
    "Sid": "Statement1",
    "Effect": "Allow",
    "Principal": {
        "AWS": "arn:aws:iam::xxx:role/kubefirst-worker-nodes-role-6zray_kubefirst_tech"
    },
    "Action": "sts:AssumeRole"
}
```
After this insertion, kubefirst init worked as expected.

marinhojess:
By removing the `currentCommand.MarkFlagRequired("profile")` line in the cmd/init.go file it's possible to init kubefirst without informing the aws profile.

marinhojess:
Even after suppressing the profile in kubefirst init, kubefirst create still exports the variable AWS_PROFILE, as shown in the following print:

![image](https://user-images.githubusercontent.com/19623668/191092825-f19cf4bd-924e-4a88-919e-4a179b275836.png)

6za:
@marinhojess is this still an issue?

PR #522, opened 2022-10-06
Remove aws profile

marinhojess:
To ensure that terraform will authenticate with role instead profile, we had to create a function to treat both cases.


marinhojess:
#406

6za (approved):
lgtm

6za (dismissed):

6za (approved):

PR #528, opened 2022-10-07
Fix #514

### LINKED ISSUE #514: Feature Parity - Gitlab installs post 1.9.0 not creating default users
# Overview
image

- [x] Github install create default admin user 
- [ ] Gitlab install create default admin user

converge:
this PR:

- add a conditional to Terraform users apply step
- this fix the issue that the Terraform user apply was applying for GitHub only

closes #514


6za:
_Reorganizing  functions places, makes harder to track changes, increase chances of something to be lost. As we are not enforcing order rules, maybe we could avoid to "re-org" the code on simple function internals changes PRs, but anyway, what is done is done. Just maybe, we should try to avoid too much churn in the code to make easier on the fellow reviewer._

internal/terraform/terraform.go
```diff
@@ -333,3 +289,71 @@ func OutputSingleValue(dryRun bool, directory, tfEntrypoint, outputName string)
...
+		return errors.New("git provider not provided, skipping terraform apply")
+	}
+
+	log.Println("Executing ApplyUsersTerraform")
+
+	//* AWS_SDK_LOAD_CONFIG=1
+	//* https://registry.terraform.io/providers/hashicorp/aws/2.34.0/docs#shared-credentials-file
+	envs := map[string]string{}
+
+	if gitProvider == "github" {
```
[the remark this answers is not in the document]
converge: I believe it would be better to rely on gitlab domain to check if gitlab is available or not. if we remove github dependency some day, gitlab will be wired to a github value.

converge:
> Reorganizing functions places, makes harder to track changes, increase chances of something to be lost. As we are not enforcing order rules, maybe we could avoid to "re-org" the code on simple function internals changes PRs, but anyway, what is done is done. Just maybe, we should try to avoid too much churn in the code to make easier on the fellow reviewer.

I agree, that was my bad, I moved it to the bottom while resolving a merge conflict.

internal/terraform/terraform.go
```diff
@@ -333,3 +289,71 @@ func OutputSingleValue(dryRun bool, directory, tfEntrypoint, outputName string)
...
+		envs["GITHUB_TOKEN"] = os.Getenv("GITHUB_AUTH_TOKEN")
+		envs["GITHUB_OWNER"] = viper.GetString("github.owner")
+	} else if gitProvider == "gitlab" {
+		envs["GITLAB_TOKEN"] = viper.GetString("gitlab.token")
+		envs["TF_VAR_initial_password"] = viper.GetString("botpassword")
```
6za, line 320: We are setting this property 2x on this function: 
- envs["TF_VAR_initial_password"]
Lines: 320 and 332.. 

I think we want to keep only line `332` as both modes will support the bot password, I believe. @johndietz  can confirm on that.

internal/terraform/terraform.go
```diff
@@ -333,3 +289,71 @@ func OutputSingleValue(dryRun bool, directory, tfEntrypoint, outputName string)
...
+
+	if viper.GetBool("create.terraformapplied.users") || dryRun {
+		log.Println("skipping: ApplyUsersTerraform")
+		return nil
+	}
+
+	if len(gitProvider) == 0 {
+		return errors.New("git provider not provided, skipping terraform apply")
+	}
+
```
6za, line 308: I am reflecting if this is an event that we should break the installation. We shouldn't have reached this far on an installation without a `gitProvider`. That is why we were using the `bool` flavor to avoid this check. 

I double-checked, it seems before a flavor creating we now set the `git.mode` preventing this to be empty: https://github.com/kubefirst/kubefirst/blob/main/cmd/create.go#L78

6za (approved):
I am approving assuming we already tested an `github` and `gitlab` installation using this branch. 

If we didn't executed a full install, please execute this installs and ensure both `github` and `gitlab` have the kbot user added on vault as described on the issue. 

https://github.com/kubefirst/kubefirst/issues/514

internal/terraform/terraform.go
johndietz: yep agreed it'll be the bot user on both git providers now and init password will be implemented on both sides

converge:
comments addressed here https://github.com/kubefirst/kubefirst/pull/545

Issue #501, opened 2022-10-04
kubefirst on the localhost


johndietz:
We're looking for a way to provide a kubefirst platform experience run entirely on the localhost. This is architecturally different from a cloud kubernetes environment in a number of different ways:
- the localhost has no native ingress available so the localhost story might need a local k8s [distro that supports that](https://k3d.io/v5.4.6/usage/exposing_services/)
- memory and cpu availability on a localhost will be significantly less available and we may need to make some choices about what pieces we deliver
- vault backend with dynamo and kms will need accommodations
- s3 storage for terraform state, argo artifacts, and chartmuseum helm charts will need accommodations
- ecr registry for containers will need accommodations

Today (1.9) many of our apps intercommunicate over the ingress' dns, and almost all of those should pivot to internal cluster references. This is a permanent change and not associated with this localhost feature, but doing so will help this localhost feature not depend on ingress so heavily.

The four most important things to be able to demonstrate from the localhost story are:
- metaphor gitops delivery with workflows
- atlantis automation of terraform
- vault secret store
- telemetry of a cluster create event

6za:
Exploring `localhost` experience using k3d:

```bash 
wget -q -O - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=v5.4.6 bash
```

Initial thoughts: 
- This would be a "lite" experience of kubefirst. 
- It would have the basic to show keys elements of kubefirst, but will rely on some local adaptations to make it work. 
- [Persistence](https://dev.to/bbende/k3s-on-raspberry-pi-volumes-and-storage-1om5) we may try to replace s3 buckets PVC for local folders on host
- It would not be a production ready version, mostly an demo experience that is operational to help developers to create metaphor like applications with a set of devops experience
- It initial version will be based on a fork of the `github` version of kubefirst to save resources on the git repo. 
- For webhook, something like [ngrok](https://ngrok.com/) may works for exposing the cluster to web-events with ephemeral URLs.

6za:
The "lite" version.. "kubefirst-step" :) 
- Would include only 1 metaphor, the original one in nodejs. 
- We would have Vault as secret manager 
> Vault setup may need to be harmonized to be setup for local use, no KMS or DynamoDB
- We would have Vault as IDP
- ArgoCD + Argo workflows 
- External-secrets 
- We wouldn't need the lets encrypt tooling as most of routes would be internal
> cert-issuer, cert-manager, .. ingress-nginx?
[...]

PR #535, opened 2022-10-08
Exploring localhost mope setup

6za:
Changes:
- Re-org validation rules
- Add test scenarios for init runs
- Update `init` bars setup

Adds:
- #501 

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -3,6 +3,12 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"log"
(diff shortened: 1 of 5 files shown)


6za:
As a general idea, `github + aws` and `github + k3d` are meant to be a very similar experience, initial draft:
image

Later if it works, we may want to converge both installation flows. 

For now we will fork to experiment and be able to break.

pagottoo (approved):
lgtm

Issue #266, opened 2022-08-24
1.9.x - Create Automation to test Gitlab Flow


6za:
# Overview

Create a an automation that executes the following commands in sequence:
> Any exit 1 is an signal of error.
```bash
# init
go run . init \
  --profile default \
  --cloud aws \
  --region us-east-1 \
  --aws-nodes-spot \
  --cluster-name cluster_name \
  --s3-suffix cicd01 \
  --admin-email email@host.io \
  --hosted-zone-name host.com

# create cluster
go run . cluster create

# backup certs
go run . backupSSL

# destroy cluster
go run . destroy 
go run . clean --destroy-buckets --destroy-confirm
go run . action remove-github # if github
```

# Notes

After destruction, ensure that: 
- VPC is removed (by tag)
- EC2/Load Balancer is removed (by tag)
- S3 Buckets are removed 
- KMS Alias is removed
- EKS Cluster is removed
- Volumes are removed for the EC2 Instances 
- *Routes are removed from Route53*-- # destroy doesn't delete routes

6za:
Although, destroy don't remove the routes, we may want to automate at  CI/Cd to remove it. To help speed the next run of the test. To make resolution of names faster.

converge:
it might be good to have solve it separated https://github.com/kubefirst/kubefirst/issues/267

converge:
initial test if there is resource

use:
—s3-suffix
—aws-nodes-sport

between create → destroy
backup ~/.k1

in case of destroy error:
 1 - apply terraform destroy (tfstate) -> based on backup
 2 - 2. API call to kill left overs resources (delete all tag based) -> based on backup

6za:
`—aws-nodes-spot` - right?

6za:
Summarizing my notes from the discussion:(this is not the solution is notes for a future improvement)
- It may exist a step 0, that knows every assets from the family of resources we care, check `notes` of the ticket description. This step has a binary on which run list operations againts the API of resources like buckets/kms/ec2/vpcs...   this will produce a file, probably a yaml of what exist today. This file is stored for a later step. 
> output: state_of_cloud.yaml 
- step 1: executes init
- step 2: executes create
- step 3: Validate if all the required elements expected were created as desired. At this step the binary of step zero could be run again, to check state state_of_cloud_post_create.yaml
[...]

PR #542, opened 2022-10-10
Insert yaml

marinhojess:
Implementing flag to install CI on gitlab flavor


marinhojess:
related issue: #266

internal/ciTools/ci.go
```diff
@@ -60,3 +67,24 @@ func SedBucketName(old, new string) error {
 
 	return nil
 }
+
+func DetokenizeCI(old, new, ciLocation string) {
```
converge, line 71: feel free to follow if it makes sense for you.

returning error in functions is nice since we can easily create unit tests for them. test will be like:
```
call DetokenizeCI(...)
if return error, fail, or you validate the expected error and so on.

logging is great, but I would log them outside in the function caller, and let the function only return the error. *ps: in some cases, is nice to log inside the function.

example:
err := DetokenizeCI(...)
if err != nil { log.Println(err) }

pagottoo (approved):

6za (dismissed):
lgtm

converge (dismissed):

pagottoo (approved):
lgtm

johndietz (approved):

PR #564, opened 2022-10-14
chore: break if k1 folder is present before init

### LINKED ISSUE #550: break kubefirst init if the kubefirst folder has content
Case:
Sometimes we forget to run `kubefirst clean` between installations, so when the `kubefirst init` cmd is rerun, the process fails; this is a bad experience.

Proposal:
If the folder of kubefirst has content, we should break the initialization command as soon as possible and inform the user about it.
We should warn the user that this content can be about a previous installation or is the `trash` of an unsuccessful installation and instruct to run the `kubefirst clean`;

converge:
closes #550 

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -91,13 +91,13 @@ validated and configured.`,
 
 		progressPrinter.SetupProgress(progressPrinter.TotalOfTrackers(), globalFlags.SilentMode)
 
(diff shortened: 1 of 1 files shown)


marinhojess (approved):

pagottoo (approved):
lgtm

PR #612, opened 2022-10-24
feat: disable github login step

converge:


jarededwards (approved):

converge:
sounds like there is another easy way to bypass our current issue, pls dont merge it yet

jarededwards:
> sounds like there is another easy way to bypass our current issue, pls dont merge it yet

i just pushed a quick change commenting out the flag value not used
error:
```# github.com/kubefirst/kubefirst/cmd
cmd/init.go:43:3: cloudValue declared but not used```

converge:
we found a better way to solve it. now GitHub login step will handler the GitHub token issue step

PR #635, opened 2022-11-02
Rewire local

converge:
This PR:

- keeps old local.go command
- adds a new local.go command
- structure the three different installation moment into functions (pre install, install, post install)
- use single source of truth for flags
- organize local into local package
- remove the conditionals to check if its local install or not

### DIFF
--- a/cmd/local/local.go
+++ b/cmd/local/local.go
@@ -1,0 +1,700 @@
+package local
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/kubefirst/kubefirst/configs"
+	"github.com/kubefirst/kubefirst/internal/argocd"
+	"github.com/kubefirst/kubefirst/internal/domain"
+	"github.com/kubefirst/kubefirst/internal/downloadManager"
+	"github.com/kubefirst/kubefirst/internal/gitClient"
+	"github.com/kubefirst/kubefirst/internal/githubWrapper"
+	"github.com/kubefirst/kubefirst/internal/handlers"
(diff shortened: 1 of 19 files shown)


cmd/local/local.go
```diff
@@ -0,0 +1,483 @@
...
+	localCmd.Flags().BoolVar(&dryRun, "dry-run", false, "set to dry-run mode, no changes done on cloud provider selected")
+	localCmd.Flags().BoolVar(&silentMode, "silent", false, "enable silentMode mode will make the UI return less content to the screen")
+	// todo: get it from GH token , use it for console
+	localCmd.Flags().StringVar(&adminEmail, "admin-email", "", "the email address for the administrator as well as for lets-encrypt certificate emails")
+
+	localCmd.Flags().StringVar(&metaphorBranch, "metaphor-branch", "main", "metaphro application branch")
+	localCmd.Flags().StringVar(&gitOpsBranch, "gitops-branch", "main", "version/branch used on git clone - former: version-gitops flag")
+	localCmd.Flags().StringVar(&gitOpsRepo, "gitops-repo", "gitops", "")
+
+	localCmd.Flags().BoolVar(&enableConsole, "enable-console", true, "If hand-off screen will be presented on a browser UI")
```
6za, line 65: this will disable the support of the config file, right?

6za: Or do we plan to mix global vars with the current read config logic?

6za (approved):

6za:
let merge and see.

cmd/local/local.go
converge: I'll tackle that after merging

PR #636, opened 2022-11-02
feat: speed up download step with

converge:
#### command:
`go run . init --gitops-branch main --metaphor-branch main --github-owner org-k1-converge-2 --github-user converge --cloud k3d`

### main branch download
Download dependencies         ... done! [3 in 15.751s]

### this PR download
Download dependencies         ... done! [3 in 3.154s]


6za (changes requested):
I was not able to find a mechanism to handle errors when one of the artifacts fail to download. 

This could produce an installer that never finishes(forever stuck) or proceed missing a binary.

6za:
Threads/go-func are good, but we still need to keep handling errors.

6za:
We returned errors before, and this change is undoing this, unless I am missing something. 
image

converge:
1. timeout handling I wouldn’t do it now, since it will require context and channels to control , and it will require more time for development.

2. error handling is actually a good point, I didnt test the non happy path.

PR updated, if it fails, it panic and block the progress.

6za (approved):
lgtm

PR #645, opened 2022-11-03
feat: use viper standard for command inputs

converge:

### DIFF
--- a/configs/viperConfig.go
+++ b/configs/viperConfig.go
@@ -1,0 +1,99 @@
+package configs
+
+// code from: https://github.com/carolynvs/stingoftheviper
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"strings"
+)
+
+// it follows Viper precedence:
+// 1. explicit call to Set
+// 2. flag
+// 3. env
+// 4. config
+// 5. key/value store
+// 6. default
+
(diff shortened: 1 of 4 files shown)


configs/viperConfig.go
```diff
@@ -0,0 +1,99 @@
...
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		// Environment variables can't have dashes in them, so bind them to their equivalent
+		// keys with underscores, e.g. --favorite-color to STING_FAVORITE_COLOR
+		v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+
+		// Apply the viper config value to the flag when the flag is not set and viper has a value
+		if !f.Changed && v.IsSet(f.Name) {
+			val := v.Get(f.Name)
+			cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val))
+		}
```
6za, line 97: Replacing a custom logic by another. lol

6za (approved):

configs/viperConfig.go
converge: I would be happy if we can replace few clean lines of code, instead of using all this logic that tried to do the same https://github.com/kubefirst/kubefirst/tree/main/internal/flagset

let's try to follow the project code of conduct, commenting "lol" is aggressive and disrespectful with the other colleges. https://github.com/kubefirst/kubefirst/blob/main/CODE_OF_CONDUCT.md

PR #685, opened 2022-11-08
Fix 664

6za:
Fix missing addons on local


6za:
hardcoded to only support github by request.

PR #727, opened 2022-11-14
fix: Remove ELB

marinhojess:

The load balancer and its security group are not being removed with kubefirst destroy, which causes the destroy process to fail when trying to remove the VPC from the cluster.

To mitigate this issue, functions were developed to remove the security group and load balancer (based on the cluster name tag) to proceed with the VPC destroy.

### DIFF
--- a/internal/terraform/terraform.go
+++ b/internal/terraform/terraform.go
@@ -4,15 +4,17 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
 	"github.com/kubefirst/kubefirst/configs"
 	"github.com/kubefirst/kubefirst/internal/aws"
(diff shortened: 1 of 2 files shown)


internal/terraform/terraform.go
```diff
@@ -184,6 +186,12 @@ func DestroyBaseTerraform(skipBaseTerraform bool) {
 			envs["TF_VAR_capacity_type"] = "SPOT"
 		}
 
+		err = aws.DestroyLoadBalancer(viper.GetString("cluster-name"))
+		if err != nil {
```
converge, line 190: not sure what we want here, double checking... it will break/stop the destroy command if it's not possible to delete the load balancer, is that what we want?

marinhojess: I think so, because the destroy will fail anyway if load balancer is not successfully removed, the vpc has dependencies with the load balancer and the terraform breaks.

converge (approved):

PR #774, opened 2022-11-22
refactor: initial argocd config, split it into local and cloud

converge:

### DIFF
--- a/pkg/keys.go
+++ b/pkg/keys.go
@@ -6,13 +6,13 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"github.com/kubefirst/kubefirst/configs"
 	"log"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/keys.go
```diff
@@ -58,26 +57,32 @@ func CreateSshKeyPair() {
...
-	if err != nil {
-		log.Panicf("error: could not write argocd-init-values.yaml %s", err)
+
+	// todo: break it into smaller function
+	if viper.GetString("gitprovider") != CloudK3d {
+
+		config := configs.ReadConfig()
+		privateKey := viper.GetString("botprivatekey")
+
+		var argocdInitValuesYaml = []byte(fmt.Sprintf(`
```
pagottoo, line 67: if I am not wrong, this alignment will broke, we need to keep aligned all to left

pagottoo (changes requested):

marinhojess (approved):

Issue #818, opened 2022-12-02
local - when using branch is producing a main-less repo. 


6za:
# Overview

Local, when build from branch is creating a mainless repo:
![image](https://user-images.githubusercontent.com/53096417/205368225-95469f3f-4dc6-468b-ba22-6dc83b2e3720.png)

Registry should be based on `main` to match cluster version.

6za:
Created using: 
```bash 
go build -ldflags="-X github.com/kubefirst/kubefirst/configs.K1Version=1.10.8" .
./kubefirst local --gitops-branch feat_813 --skip-metaphor
```

6za:
Reviewing the new local clone/checkout logic.. it seems to be missing this logic to add main from branches/tags: https://github.com/kubefirst/kubefirst/blob/1_11_release/internal/gitClient/git.go#L273

Local seems to be producing main-less repos at github.

6za:
It seems to missing some logic here, to address it:
- https://github.com/kubefirst/kubefirst/blob/1_11_release/internal/gitClient/git.go#L528
- https://github.com/kubefirst/kubefirst/blob/1_11_release/internal/gitClient/git.go#L510

I am not sure if this is by design, but it seems to be odd, to be `main-less` the repo. I believe the 1st branch will be made "HEAD" on this scenarios.

6za:
Merged to Release 1.11 branch

PR #820, opened 2022-12-05
feat: enable atlantis apply for non-main branches

converge:


6za:
I believe we should favor to have always a main, as delivered at #818. As for gitops POV we need to have a Desired State branch and that by convention is main today.

PR #825, opened 2022-12-06
feat: add redirect to https from ingress

marinhojess:
- Change gitops-branch to add-ingress-localhost
- Remove certificate secret creation from postrun and move it to local.go
- Move the installation of argocd ingressroute.yaml to local.go and create a new ingress to force redirect connections from argocd to https
- Fix vault endpoint to use https instead of http
- Force creation of kubefirst and minio namespaces at start of installation, thus allowing certificate secrets to be created at beginning
- Remove some unnecessary annotations from argocd ingress

### DIFF
--- a/cmd/local/local.go
+++ b/cmd/local/local.go
@@ -2,9 +2,10 @@ package local
 
 import (
 	"fmt"
-	"github.com/rs/zerolog/log"
 	"sync"
 	"time"
+
+	"github.com/rs/zerolog/log"
 
 	"github.com/kubefirst/kubefirst/configs"
 	"github.com/kubefirst/kubefirst/internal/wrappers"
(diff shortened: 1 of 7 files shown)


cmd/local/local.go
```diff
@@ -419,6 +426,16 @@ func runLocal(cmd *cobra.Command, args []string) error {
 		progressPrinter.IncrementTracker("step-telemetry", 1)
 	}
 
+	_, _, err = pkg.ExecShellReturnStrings(config.KubectlClientPath, "--kubeconfig", config.KubeConfigPath, "-n", "argocd", "apply", "-f", fmt.Sprintf("%s/gitops/ingressroute.yaml", config.K1FolderPath))
+
+	if err != nil {
+
+		log.Printf("failed to create ingress route to argocd: %s", err)
```
converge, line 433: I'll advocate a bit to use zerolog for logging xD

cmd/local/local.go
```diff
@@ -419,6 +426,16 @@ func runLocal(cmd *cobra.Command, args []string) error {
...
+
+	}
+
+	_, _, _ = pkg.ExecShellReturnStrings(config.KubectlClientPath, "--kubeconfig", config.KubeConfigPath, "-n", "argocd", "delete", "ingress", "argocd-server")
```
converge, line 437: the 3rd return value is an error, it would be nice to log it, like:
```
log.Error().Err(err).Msg("")
```

it will generate a red log line with error level

cmd/local/local.go
```diff
@@ -65,8 +65,9 @@ func NewCommand() *cobra.Command {
 	// todo: UPDATE IT BEFORE MERGING
 	// todo: UPDATE IT BEFORE MERGING
 	// todo: UPDATE IT BEFORE MERGING
-	localCmd.Flags().StringVar(&gitOpsBranch, "gitops-branch", "add-ingress-localhost", "version/branch used on git clone")
```
converge: I guess we should keep this hardcoded reference until we merge that gitops-template branch, wdyt?

marinhojess: hey @converge, with the last PR that @6za did, now we just need to specify gitops-branch when calling the local installation if we want to test a different branch, like following:

`./kubefirst local --gitops-branch add-ingress-localhost`

converge: not sure if I got it, we can do that with any flag available since the beginning of the command...

no preference, we can use --gitops-branch as well

converge (approved):

PR #929, opened 2022-12-15
feat: Add retry logic to create ngrok tunnel

marinhojess:
- Add retry logic to create ngrok tunnel
- Will try to create the ngrok tunnel 3 more times if the first attempt fails

### DIFF
--- a/pkg/ngrok.go
+++ b/pkg/ngrok.go
@@ -3,9 +3,11 @@ package pkg
 import (
 	"context"
 	"fmt"
-	"github.com/rs/zerolog/log"
 	"io"
 	"net"
+	"time"
+
+	"github.com/rs/zerolog/log"
 
 	"github.com/ngrok/ngrok-go"
 	"github.com/ngrok/ngrok-go/config"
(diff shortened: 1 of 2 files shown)


6za (approved):
Loved the function pointer.

pkg/ngrok.go
```diff
@@ -41,6 +53,33 @@ func RunNgrok(ctx context.Context) {
 	}
 }
 
+func init() {
```
jarededwards, line 56:
```suggestion
func init() {
```
is this function being used? i didnt see a reference in your code?

jarededwards: how often is this failing? with all the hard work of cleaning up logs and updating the logger to produce better output is this going to produce 10 times of unnecessary noise?

marinhojess: @jarededwards you are right, forgot to remove, done! thanks!

6za (approved):

pagottoo (approved):
lgtm

PR #944, opened 2022-12-16
fix: Include env var for region

marinhojess:
Terraform provider needs to know which region to use to authenticate to AWS, if the region doesn't exist in the AWS shared file, it must be informed in another way.

An alternative is to include the region informed by the installer flag as an AWS_REGION environment variable, on this way terraform provider will be able to identify the correct version through the flag.


pagottoo (approved):
lgtm

Issue #926, opened 2022-12-15
Cloud install  - kubeconfig been added to gitops repo, it shouldn't


6za:
![image](https://user-images.githubusercontent.com/53096417/207876724-9c115575-0cc0-4c52-84f1-edf6951e216e.png)

6za:
File also present on gitlab:

6za:
File filter working: 
```bash 
8:39PM DBG gitlab.go:691 > checking file 'registry/aws-auth.yaml'
8:39PM DBG gitlab.go:690 > the file is terraform/base/kubeconfig the status is 63
8:39PM DBG gitlab.go:691 > checking file 'terraform/base/kubeconfig'
8:39PM DBG fileFilter.go:18 > file not included on commit: 'terraform/base/kubeconfig'
8:39PM DBG gitlab.go:690 > the file is validation/tokens.md the status is 77
8:39PM DBG gitlab.go:691 > checking file 'validation/tokens.md'
```

6za:
Added a way on #953  to filter files before to add to commit to main.

Issue #896, opened 2022-12-14
[main] 1.11 Old logs on Init


6za:
As we migrated the logs on 1.11 this are instances where we still have the older log format. 

Init: 

```bash 

LOG: 2022/12/14 12:46:02.665428 /home/developer/app/cmd/init.go:247: creating buckets for state and artifacts
LOG: 2022/12/14 12:46:02.667471 /home/developer/app/internal/aws/aws.go:113: creating state-store bucket k1-state-store-suppresed-gh-12-14
LOG: 2022/12/14 12:46:02.667955 /home/developer/app/internal/aws/aws.go:116: region is  us-east-1
LOG: 2022/12/14 12:46:03.793877 /home/developer/app/internal/aws/aws.go:546: tagging bucket... k1-state-store-suppresed-gh-12-14:suppresed
LOG: 2022/12/14 12:46:04.605437 /home/developer/app/internal/aws/aws.go:582: Bucket: k1-state-store-suppresed-gh-12-14 tagged successfully
LOG: 2022/12/14 12:46:04.609725 /home/developer/app/internal/aws/aws.go:160: bucket k1-state-store-suppresed-gh-12-14 exists
LOG: 2022/12/14 12:46:04.610431 /home/developer/app/internal/aws/aws.go:113: creating argo-artifacts bucket k1-argo-artifacts-suppresed-gh-12-14
LOG: 2022/12/14 12:46:04.611089 /home/developer/app/internal/aws/aws.go:116: region is  us-east-1
LOG: 2022/12/14 12:46:05.785137 /home/developer/app/internal/aws/aws.go:546: tagging bucket... k1-argo-artifacts-suppresed-gh-12-14:suppresed
LOG: 2022/12/14 12:46:06.645437 /home/developer/app/internal/aws/aws.go:582: Bucket: k1-argo-artifacts-suppresed-gh-12-14 tagged successfully
LOG: 2022/12/14 12:46:06.648665 /home/developer/app/internal/aws/aws.go:160: bucket k1-argo-artifacts-suppresed-gh-12-14 exists
LOG: 2022/12/14 12:46:06.649135 /home/developer/app/internal/aws/aws.go:113: creating gitlab-backup bucket k1-gitlab-backup-suppresed-gh-12-14
LOG: 2022/12/14 12:46:06.649665 /home/developer/app/internal/aws/aws.go:116: region is  us-east-1
[...]

Issue #363, opened 2022-09-12
Prevent logs to write `credentials` from installs


6za:
# Overview

We have several log messages, that need to be disabled or protected under a flag, to prevent log of some key credentials: 

## Examples: 
```bash 
LOG: 2022/09/12 13:46:34.422067 /home/developer/app/cmd/createUtils.go:243: {"keys":[],"keys_base64":[],"recovery_keys":
```

6za:
Mitigated on #953.

6za:
No creds were found on 1.11 effort after post #896 to be completed.

PR #953, opened 2022-12-16
fix: #926 and tackle #896

6za:
Solves: 
- #926 

Mitigates:
- #896 
- #363

Add a basic support for future tooling to filter files in commits based on rules.


6za:
I kept the log level and error at the same to avoid break behaviors.

pagottoo (approved):
lgtm

PR #973, opened 2022-12-21
fix local FAQ URL

fharper:
Signed-off-by: Frédéric Harper <hi@fred.dev>


fharper:
Won't need that as I'm merging the FAQ

PR #1025, opened 2022-12-29
smart_download

pagottoo:
With this PR, we will download the tools (terraform/helm/k3d/kubectl) only if they do not exist in the tools folder.
This will save network traffic, rate limits, and time, avoiding re-downloading tools every clean/init command.
This will be very useful at conferences, hotels, and mobile connections where the internet could be poor.
We introduce a new flag on clean cmd, called `--preserve-tools`, to keep the `tools` folder and delete all the other resources on folders. If the tools folder doesn't exist, we will download it again the next time cmd init is executed, but if the tools exist we will skip the re-download.


fharper (dismissed):
docs LGTM

6za (changes requested):
as discussed on around,

Two major points to improve and mitigate any support concerns that may rise from this changes: 

- [x] We can't trust the files versions in the `tools` folder between runs(people may be testing different releases), so if we plan to keep the files, we need to have some metadata to confirm they are of the version we desire. For a more simple mechanism, I suggest some sort of `.cache` file on the tools dir, per binary of interest. 

- [ ] `clean` default behavior must be to clean all, and we could have something like `--preserve-tools` to allow more advanced users to keep these binaries, but without cache and and a better way to clean all folders created(more generic approach), it is better to just clean all files in most of the cases. 


My two points are related to reduce the chance of mixed versions of binaries or old folders breaking re-runs by users not too familiar with the internals of the tools.

6za:
This is a great direction towards a better user experience when running with limited connection scenarios. Great work!

pagottoo:
> * We can't trust the files versions in the `tools` folder between runs(people may be testing different releases), so if we plan to keep the files, we need to have some metadata to confirm they are of the version we desire. For a more simple mechanism, I suggest some sort of `.cache` file on the tools dir, per binary of interest.

To solve this, I used an approach to check directly on the tool the version of binary using commands like `helm version` to check against the version defined on `config`. If the file exists and the version match, we skip the download, if the file exists, and the version doesn't match, we show an error message recommending the user to run `kubefirst clean`, and if the file does not exist, we download.

> * clean default behavior must be to clean all, and we could have something like --preserve-tools to allow more advanced users to keep these binaries, but without cache and and a better way to clean all folders created(more generic approach), it is better to just clean all files in most of the cases.

Suggestion accepted, changed the default behavior to always clean the tools folder, and if desired, informed by the flag `--preserve-tools` we skip the removal of the tools folder.

cmd/clean.go
```diff
@@ -58,22 +59,37 @@ re-create Kubefirst base files. To destroy cloud resources you need to specify a
...
+
+		if preserveTools {
+			// delete gitops path and argo init values - caching tools to avoid re-download
+			err = os.RemoveAll(config.GitOpsRepoPath)
+			if err != nil {
+				return fmt.Errorf("unable to delete %q folder, error is: %s", config.GitOpsRepoPath, err)
+			}
+			err = os.Remove(config.ArgoCDInitValuesYamlPath)
+			if err != nil {
+				return fmt.Errorf("unable to delete %q file, error is: ", err)
```
6za, line 75: don't forget the metaphor repos...

docs/tooling/kubefirst-cli.md
```diff
@@ -26,7 +26,8 @@ kubefirst clean [flags]
 Flags:
 --destroy-buckets   destroy buckets created by init cmd
 --destroy-confirm   confirm destroy operation (to be used during automation process)
--h, --help              help for clean
+-h, --help          help for clean
+--preserve-tools    preserve all tools downloaded (avoid re-downloading)
```
fharper, line 30: preserve all downloaded tools (avoid re-downloading)

fharper (changes requested):
small change in the CLI docs, worth changing in the CLI also. Not a big deal though

docs/tooling/kubefirst-cli.md
pagottoo: tks

fharper (approved):
docs LGTM

6za (dismissed):

cmd/local/prerun.go
```diff
@@ -34,8 +34,6 @@ func validateLocal(cmd *cobra.Command, args []string) error {
 
 	config := configs.ReadConfig()
 
-	log.Info().Msg("sending init started metric")
```
converge, line 37: good call! we have two logs for the same thing. 

when debugging telemetry, it's nice to read the logs to check if the telemetry send function was called, wdyt to update the log message from
```
pkg.InformUser("Sending installation telemetry", silentMode)
```
to
```
pkg.InformUser("Sending kubefirst.init.started installation telemetry", silentMode)
```

p.s: we can update it on a separated issue/PR to also update the other references, there is also constants we can use for telemetry names.

pkg/helpers_test.go
```diff
@@ -127,7 +128,7 @@ func TestValidateK1Folder(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	_, err = os.CreateTemp(populatedTempFolder, "test.ext")
+	_, err = os.Create(fmt.Sprintf("%s/%s", populatedTempFolder, "argocd-init-values.yaml"))
```
converge, line 131: `os.CreateTemp` is preferable for unit tests, `os.Create` creates files in the host machine and we will be leftovers in the host machine.

`os.CreateTemp` will create the resource, and self delete it when the test finish.

converge (changes requested):
nice idea! just minor requests.

the clean command description says we do re-create the ~/k1 folder, but the PR is removing the re-creation part.

pkg/helpers_test.go
6za: Do we have confirmation we are migrating to `CreateTemp`, I don't think this PR is changing this behavior. Would this be a blocking change?

converge: there is actually no migration, `os.CreateTemp` is the current code.

In Go, the os.Create function is used to create a new file or truncate an existing one. However, when writing unit tests, it's generally a good idea to avoid modifying the file system. Instead, you can use `os.CreateTemp` function to create a temporary file that is automatically cleaned up after the test is finished. This allows you to isolate your tests and avoid any unintended side effects.

Additionally, as Go's os.Create function opens the file with mode 0666 (before umask), in general it may lead to security issues if the test is run in a shared environment, testing.TempFile creates a file with mode 0600, which is more secure.

6za: https://github.com/kubefirst/kubefirst/issues/1097

converge: > Maybe use the existing ticket to track this change: - #1097

will do

converge: synced with Thiago in slack as well, I'll take it from here in the ticket about

cmd/local/prerun.go
converge: we will have telemetry update this week, this can wait.

6za (approved):

PR #1056, opened 2023-01-05
feat: format log information

converge:
before:
Screenshot 2023-01-05 at 10 53 25

after:
Screenshot 2023-01-05 at 10 53 03

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


6za:
The new placing is kinda weird, we may want to break the line to be easier to copy and paste, the original issue was that was hard to find how to read the logs. 

It was too discrete.

PR #1057, opened 2023-01-05
docs: remove experimental

converge:
As a user, it doesn't add much value to my workflow / experience with Kubefirst.

What is missing, is in the CONTRIBUTING.md, how to contribute with the `cloud` experience, we are covering `local` only, which is good. Covering only `local` make the document right to the point, less cognitive effort to start contributing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,13 +174,6 @@ Kubefirst provides extra tooling for handling the provisioning work.
 
 ![kubefirst provisioning diagram](/images/provisioning.png)
 
-
-## Experimental Mode (we discourage it)
-
-If you want to run the main branch (`go run . init ...`, for example), you need to inform the version you are trying to use. To achieve this, you need to pass the `ldflags` in your command: `go run -ldflags="-X github.com/kubefirst/kubefirst/configs.K1Version=1.9.3" . version`. 
-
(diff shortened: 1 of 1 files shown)


converge:
> Make sense, but can we be sure to add the flag information in https://github.com/kubefirst/kubefirst/blob/main/CONTRIBUTING.md#getting-started-with-the-code please?
> 

We actually don't need ldflags to run the `local` and `cloud` installs. But I just found a bug on `local` , it's not accepting the no flags state anymore. No flags state: `go run . local` 

> The latter is not final. I'll create an issue to add information about using `main` for the cloud also. I think it needs to be there for contributors to be able to contribute not just on `local`.

Agreed

fharper:
Sounds good, I didn't even know you needed the flags at first, I run it like I wrote it in `CONTRIBUTING.md`.

So LGTM without further changes elsewhere than.

converge:
> Sounds good, I didn't even know you needed the flags at first, I run it like I wrote it in `CONTRIBUTING.md`.

True, we're failing a bit on that one, the flag wasn't required, but now it's required again.

> So LGTM without further changes elsewhere than.

PR #1066, opened 2023-01-06
docs: add new faq items

converge:


fharper (approved):
LGTM, good job, thanks

The only suggestion is that if we have an issue for the known bug downloading Terraform providers, we should link to it.

converge:
> LGTM, good job, thanks
> 
> The only suggestion is that if we have an issue for the known bug downloading Terraform providers, we should link to it.

good idea! I just added it

fharper (approved):
LGTM

PR #1069, opened 2023-01-09
Fix 1038

### LINKED ISSUE #1038: [WSL][Localhost] Error when pressing ENTER in localhost installation
### **Description**

Error when trying to open github page at the beginning of localhost installation.

### **Environment**

**kubefirst-cli golang utility version**: development
**operating system**: Linux - Ubuntu (WSL)

### **Screenshots**

image

### Steps to reproduce

Run `go run . local`

### Expected behaviour

Press Enter and open github page on browser.

converge:
this PR:
- enable Linux machines to open GH login token issue step
- uses non blocker call to open user browser

closes #1038

### DIFF
--- a/internal/handlers/github.go
+++ b/internal/handlers/github.go
@@ -90,6 +90,7 @@ func (handler GitHubHandler) AuthenticateUser() (string, error) {
 	}
 
 	if err = pkg.OpenBrowser("https://github.com/login/device"); err != nil {
+		log.Error().Msgf("error opening browser: %s", err)
 		return "", err
 	}
 
--- a/internal/handlers/github.go
+++ b/internal/handlers/github.go
@@ -46,7 +46,7 @@ func NewGitHubHandler(gitHubService *services.GitHubService) *GitHubHandler {
 // grant permission to return a valid access token.
 func (handler GitHubHandler) AuthenticateUser() (string, error) {
 
-	gitHubDeviceFlowCodeURL := pkg.GitHubLoginDeviceURL + "/code"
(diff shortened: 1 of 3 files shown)


internal/handlers/github.go
```diff
@@ -90,7 +89,8 @@ func (handler GitHubHandler) AuthenticateUser() (string, error) {
 		return "", err
 	}
 
-	if err = exec.Command("open", "https://github.com/login/device").Start(); err != nil {
+	if err = pkg.OpenBrowser("https://github.com/login/device"); err != nil {
+		log.Error().Msgf("error opening browser: %s", err)
 		return "", err
```
6za, line 94: Will this error fail the installer?

6za: For containers run, there is not "browser".. so it would prevent container runs or machine without X if this breaks the installer, if not a error that would fail the installer. I see no issues with it.

converge: the github device validation needs a browser to issue a token, containers users needs to pass the token via configuration

6za: the point is that if tis fails, I can test device token from container. I can access the link, but I can't open the browser. 

If this break the install the missing browser, we should avoid. As it would work, I used before without token. We are failing for no good reason the installer.

converge: I actually didnt add this into this PR, I just added the log.

let's move this discussion into a different topic since it's not the PR goal https://github.com/kubefirst/kubefirst/discussions/1076

6za (approved):

PR #1074, opened 2023-01-09
feat: update telemetry flag name

converge:
fyi; closes #1074 and it's part of https://github.com/kubefirst/kubefirst/discussions/1032 


converge:
Cristhian and I will work together in this one after 1.11 release, to also update the references for console UI, so that, we have a common name for CLI and console.

PR #1087, opened 2023-01-10
remove the auto assignation of docs issue to @fharper

fharper:
it is just confusing for now


converge (approved):
sounds good! fyi; the DCO failed

PR #1117, opened 2023-01-12
chore: update terraform logs to info level

### LINKED ISSUE #1109: terraform output from execution should be classified as info level log output
a recent logs refactor classifies terraform output as debug instead of info. 
a different recent change to make `info` the default log level is causing that tf output content to be omitted.
cli users will want that output by default
we will want it available for diagnosis / affirmation during triage.

AC:
- keep the default log level of info
- ensure all terraform execution detail is not omitted

converge:
closes #1109 

### DIFF
--- a/pkg/shell.go
+++ b/pkg/shell.go
@@ -40,7 +40,7 @@ func ExecShellWithVars(osvars map[string]string, command string, args ...string)
 	for k, v := range osvars {
 		os.Setenv(k, v)
 		suppressedValue := strings.Repeat("*", len(v))
-		log.Debug().Msgf(" export %s = %s", k, suppressedValue)
+		log.Info().Msgf(" export %s = %s", k, suppressedValue)
 	}
 	cmd := exec.Command(command, args...)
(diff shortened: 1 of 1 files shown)


marinhojess (approved):

pkg/shell.go
```diff
@@ -36,7 +40,7 @@ func ExecShellWithVars(osvars map[string]string, command string, args ...string)
 	for k, v := range osvars {
 		os.Setenv(k, v)
 		suppressedValue := strings.Repeat("*", len(v))
-		log.Info().Msgf(" export %s = %s", k, suppressedValue)
+		log.Debug().Msgf(" export %s = %s", k, suppressedValue)
 	}
```
6za, line 44: We need this at info level.

This allow us to know if we forgot to pass something to terraform

converge: my understanding is that if you need that for development, it's a debug data, end user won't care about it (info level).  but I'll follow your requests to unblock the PR

6za (approved):

Issue #939, opened 2022-12-16
Ngrok + atlantis / GitHub webhook not responding


converge:
at current `main` branch, Ngrok URL is created and registered, but GitHub webhook fails to reach Ngrok address, making installation final Atlantis steps to fail.

p.s: it was working before, but was broke at some point.

converge:
@pagottoo have a look at this "Problems using Ngrok" session https://www.twilio.com/blog/using-ngrok-2022 , it might be the root cause of our issues. The explanation starts at "Another more serious problem is shown in the following screenshot."

pagottoo:
Moving to triage due to the impossibility of reproduction;

converge:
I could reproduce it with:

```
go run . local --metaphor-branch main
```

converge:
adding to 1.11.1 milestone since it's still happening for some people (it works most of the time, but not always).

6za:
There is several discussion on this ticket and how hard it is to reproduce it, I would just clarify that I don't believe this issue should not block the next release, in case that still the case and we can provide evidence of its reproduction. 


It may be related to some flaky behavior of ngrok lib in use. To keep researching it is important, just to be careful to not hold next release due to the date commits, in particular to not block CIVO effort.

marinhojess:
Localhost nstallation completed successfully, even when not able to perform the atlantis plan and apply:

```
2023-01-19T14:04 INF ../../../../kubefirst/pkg/ngrok.go:50 > connection closed: <nil>
2023-01-19T14:04 INF ../../../../kubefirst/internal/githubWrapper/github.go:274 > waiting "atlantis plan" finish to proceed...
2023-01-19T14:04 INF ../../../../kubefirst/internal/githubWrapper/github.go:274 > waiting "atlantis plan" finish to proceed...
2023-01-19T14:04 INF ../../../../kubefirst/pkg/ngrok.go:50 > connection closed: <nil>
2023-01-19T14:04 INF ../../../../kubefirst/internal/githubWrapper/github.go:274 > waiting "atlantis plan" finish to proceed...
2023-01-19T14:04 INF ../../../../kubefirst/pkg/ngrok.go:50 > connection closed: <nil>
2023-01-19T14:04 INF ../../../../kubefirst/internal/githubWrapper/github.go:274 > waiting "atlantis plan" finish to proceed...
[...]

Issue #1099, opened 2023-01-11
ngrok memory leak


converge:
As a Kubefirst developer I want to handle a channel that is open for ngrok. So that, after the channel is not in use anymore, we close the running go routine and avoid memory leak.

AC1: ngrok functionality behaves the same
AC2: the opened channel is closed when the installation finish

6za:
Once the channel is closed, `atlantis` will not be usable on local. Is this the intended behavior?

converge:
> Once the channel is closed, `atlantis` will not be usable on local. Is this the intended behavior?

yes, a non closed channel is a memory leak, we can close the channel before exiting the main process/kubefirst install.

PR #1228, opened 2023-01-20
fix: fix ngrok memory leak

converge:
closes: https://github.com/kubefirst/kubefirst/issues/939 and https://github.com/kubefirst/kubefirst/issues/1099


cmd/local/postrun.go
```diff
@@ -36,14 +34,25 @@ func runPostLocal(cmd *cobra.Command, args []string) error {
...
+	//	<-sigs
+	//	wg.Done()
+	//}()
+	//wg.Wait()
+
+	// todo: testing
+	cancelContext()
+	fmt.Println("---debug---")
+	fmt.Println("context killed, waiting...")
+	fmt.Println("---debug---")
```
6za, line 52: please, prefer to use the log.debug feature to prevent this code ed-up on main, requiring us to fix later in further PR. by testing  already using the debug logger, helps to prevent his clean up prs later.

6za: That is cool we have the new logger, that prevents us to send weird console messages to our users. :)

converge: Screenshot 2023-01-20 at 11 00 37

6za (approved):

PR #1246, opened 2023-01-26
adding a couple details to local install guide

johndietz:


jarededwards:
we need to add back in the mention of collision for the `gitops` and `metaphor-frontend` repositories

jarededwards (approved):

Issue #1221, opened 2023-01-18
kubefirst version display log information


converge:
the `version` command displays unnecessary information for logs:

```
-----------
Follow your logs with:
   tail -f  /Users/converge/workspace/kubefirst/tmp/logs/log_1674048799.log

-----------


kubefirst-cli golang utility version: 1.11.0
```

**AC1:** it should display the version only

echoboomer:
@jarededwards @johndietz this issue can be closed when you have a chance. :)

jarededwards:
resolved: #1255

PR #1255, opened 2023-02-02
Exclude log helper output from specific subcommands.

echoboomer:
Addresses: https://github.com/kubefirst/kubefirst/issues/1221

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,10 +5,11 @@ package main
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


echoboomer:
This does also prevent output re: log helper when only the root command is called.

jarededwards (approved):
lgtm! thanks @echoboomer !

johndietz (approved):

Issue #1248, opened 2023-01-26
check for existence of cloud resources before allowing clean to run


jarededwards:
**What would you like to be added**:
an MVP check for existence of cloud resources before executing the `kubefirst clean` command.
initial resources to check
- [ ] eks cluster
- [ ] vpc
- [ ] github teams
- [ ] github repos

**Why is this needed**:
if a user runs `kubefirst clean` before running `kubefirst cluster destroy` they are left with manually cleaning up all the resources left in the cloud as well as github due to the configurations being lost for how the resources were created

PR #1258, opened 2023-02-03
Add check to clean command to remove cluster resources first.

echoboomer:
Signed-off-by: echoboomer <scott@echoboomer.net>

Addresses: https://github.com/kubefirst/kubefirst/issues/1248

Add a check to the `clean` command to advise a user to first run `kubefirst cluster destroy` if there are flags indicating resources have already been created. This could be automated, but I think it's important for the user to understand the implications before running.

Also changes the message in `cmd/destroy.go` returned from the `viper.GetString("cloud")` check to more plainly indicate that resources likely don't exist if you haven't already run `init` since there shouldn't be a scenario where the flag is something else or doesn't exist.

### DIFF
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -18,14 +18,31 @@ import (
 // cleanCmd removes all kubefirst resources created with the init command
 var cleanCmd = &cobra.Command{
 	Use:   "clean",
-	Short: "removes all kubefirst resources created with the init command",
-	Long: `Kubefirst creates files, folders, cloud buckets and download tools during installation at your environment. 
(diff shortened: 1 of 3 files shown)


echoboomer:
Understood if this isn't really the way we want to handle this, but in the interim, it provides the guard rails necessary to advise on the condition and prevent abandoned/orphaned resources.

jarededwards:
@echoboomer thanks for opening this one! we're in the middle of some work that has a tight deadline for our conference monday through wednesday next week ill revisit this asap! feel free to hop in the community slack if you want to talk more about this between now and then
kubefirst.io/slack

PR #1287, opened 2023-02-14
bugfix: Don't return err if dir not found during backup op.

echoboomer:


echoboomer:
```bash
2023-02-14T14:26 INF cmd/civo/civo.go:310 > checking for tls secrets to restore
2023-02-14T14:26 INF cmd/civo/civo.go:313 > open /Users/scott/.k1/ssl/your-dns.io/secrets: no such file or directory
2023-02-14T14:26 INF cmd/civo/civo.go:324 > no files found in secrets directory, continuing
```

jarededwards (approved):

PR #1292, opened 2023-02-16
feat: Add Vault bootstrap config for Civo.

echoboomer:
WIP

This should:

- Allow us to use the Vault API in the existing port-forward to initialize `vault-0`
- Utilize Pod exec (new feature) to allow us to join `vault-1` and `vault-2` to the raft cluster and unseal them
- From there, Terraform for Vault can be run to populate secrets, etc.


echoboomer:
One question I do have is what we should do with `k1_local_vault_token` when used with Civo. We may need to leave it there as a placeholder, but we'll have to figure out how to tell Atlantis to use the Secret that gets created for kube auth.

https://github.com/kubefirst/gitops-template/pull/278/files#diff-076b647de13613f378af69f2e27a55f3c0452ebb174e071d881eafe827151470R62

jarededwards (approved):

jarededwards (approved):

PR #1307, opened 2023-02-24
remove the v in front of Go version information

fharper:


fharper:
It prevents display like this

CleanShot 2023-02-24 at 13 26 30@2x

when `go version` return 

CleanShot 2023-02-24 at 13 26 37@2x

and remove nothing if it returned only numbers, since there is already a label in front of the version.

robertkeithrippetoe (approved):

PR #1328, opened 2023-03-02
bugfix: fix error without git runner creation

echoboomer:
- Fixes bug where secret name for git runner wasn't correct between the two providers.
- Fixes bug where destroy progress tracker didn't finish properly.
- Expands quota parameters checked for civo.

### DIFF
--- a/internal/civo/bootstrapSecrets.go
+++ b/internal/civo/bootstrapSecrets.go
@@ -23,14 +23,16 @@ func BootstrapCivoMgmtCluster(dryRun bool, kubeconfigPath string, gitProvider st
 	}
 
 	// Set git provider token value
-	var tokenValue, containerRegistryHost string
+	var containerRegistryHost, gitRunnerSecretName, tokenValue string
 	switch gitProvider {
 	case "github":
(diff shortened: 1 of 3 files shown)


internal/civo/bootstrapSecrets.go
```diff
@@ -143,7 +145,7 @@ func BootstrapCivoMgmtCluster(dryRun bool, kubeconfigPath string, gitProvider st
 		},
 		// git runner
 		{
-			ObjectMeta: metav1.ObjectMeta{Name: "controller-manager", Namespace: fmt.Sprintf("%s-runner", gitProvider)},
+			ObjectMeta: metav1.ObjectMeta{Name: gitRunnerSecretName, Namespace: fmt.Sprintf("%s-runner", gitProvider)},
```
jarededwards, line 148:
```suggestion
			ObjectMeta: metav1.ObjectMeta{Name: gitRunnerSecretName, Namespace: fmt.Sprintf("%s-runner", gitProvider)},
```
i think we need accommodate namespace to be `controller-manager` for github runner. maybe add another variable for each of the switches?

jarededwards (approved):

PR #1376, opened 2023-03-19
feat: kubefirst.dev local domain

jarededwards:


jarededwards:
good call, ill do a better fix before merging. we're just tinkering with the idea, might be kind of neat!

PR #1464, opened 2023-04-12
Feat vultr healthcheck

echoboomer:


fharper:
I think I fixed the kubefirst code, but I think we need to update the runtime also. It may not be the case, so feel free to revert my commit [`ef1f889` (#1464)](https://github.com/kubefirst/kubefirst/pull/1464/commits/ef1f889ad4d92833ac193b835350dfad6b5dde17). My knowledge of our code & Go is limited.

Issue #1562, opened 2023-05-29
If port 9000 is already used, fail gracefully


fharper:
### Which version of kubefirst are you using?

v2.0.8

### Which installation type?

k3d (local)

### Which distributed Git provider?

None specific

### What is the issue?

When creating a k3cd cluster, if port 9000 is used, the installation fail with an error shown in the log `not able to open port-forward error="not able to open port-forward"`, but the progress continue as if nothing happens. The installation process should stop, and display an error in the terminal. The https://github.com/kubefirst/kubefirst/issues/1561 should probably mitigate most cases, but we should still fail gracefully as the port may start to be used in between the preflight checks & the moment where MinIO need to port forward.

```
LOG: 2023/05/28 06:34:24.121353 github.com/kubefirst/runtime@v0.1.15/pkg/k8s/portForward.go:99: Namespace for PF minio
LOG: 2023/05/28 06:34:24.121610 github.com/kubefirst/runtime@v0.1.15/pkg/k8s/portForward.go:100: Name for PF minio-76cbb484b5-d78zx
2023-05-28T06:34 ERR runtime/pkg/k8s/OpenPortForwardPodWrapper/wrappers.go:58 > not able to open port-forward error="not able to open port-forward"
```

```
Running preflight checks                ... done! [5 in 7.075s]
Cloning and formatting git repositories ... done! [1 in 1m32.125s]
Applying github Terraform               ... done! [1 in 18.245s]
Pushing git repositories                ... done! [1 in 6.43s]
Creating K3d cluster                    ... done! [1 in 1m10.757s]
Bootstrapping Kubernetes resources      ... done! [2 in 7.943s]
Verifying Kubernetes cluster is ready   ... done! [3 in 24.084s]
Installing and configuring ArgoCD       ...  0.0% [......................................] [0 in 6.237422s]
Installing and configuring ArgoCD       ... done! [3 in 1m25.387s]
Configuring Vault                       ... 75.0% [############################..........] [3 in 11m17.750993s]
[#####################################################################################...] [16m30s]
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #1561, opened 2023-05-29
Preflight checks should verify if port 9000 is available


fharper:
### Which version of kubefirst are you using?

v2.0.8

### Which installation type?

k3d (local)

### Which distributed Git provider?

None specific

### What is the issue?

Since MinIO is using the port 9000, on a k3d installation, the preflight checks should check if it's available, even if it could change between the check and the port forwarding, it lower the possible issues. If not, the installation will failed.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1563, opened 2023-05-29
fix: check for port 9000 on k3d

echoboomer:
Resolves https://github.com/kubefirst/kubefirst/issues/1562 and https://github.com/kubefirst/kubefirst/issues/1561


fharper:
It fixes half of #1562. If the port wasn't in use when the create command start, but become used before MinIO needs it, it will cause the same issue looking like the cluster takes forever to create, when in fact, there's an error in the log. Maybe it's an edge case we don't want to cover, but want to be sure it's intended :)

fharper (dismissed):

PR #1753, opened 2023-08-16
fix: check new version from homebrew-core instead of k1 repo (fix #1746)

### LINKED ISSUE #1746: Upgrade CLI message should depend on Homebrew not GitHub release
### Which version of kubefirst are you using?

ALL

### Which cloud provider?

None specific

### Which installation type?

None specific

### Which distributed Git provider?

None specific

### What is the issue?

Since it take a couple of hours between a new release, and Homebrew core PR to be merged, the upgrade message should check if a newer version is available on Homebrew, and not from the kubefirst repository release.

It is needed since this messager focus on macOS user.

### Code of Conduct

fharper:


internal/common/common.go
```diff
@@ -33,17 +49,42 @@ func CheckForVersionUpdate() {
...
+		}
+
+		bodyString := string(bodyBytes)
+		for _, sentence := range strings.Split(bodyString, "\n") {
+			if strings.Contains(sentence, "url \"https://github.com/kubefirst/kubefirst/archive/refs/tags/") {
+				re := regexp.MustCompile(`.*/v(.*).tar.gz"`)
+				matches := re.FindStringSubmatch(sentence)
+				latestVersion = matches[1]
+			}
+		}
```
Claywd, line 78: I would probably skip this and regex the whole body string. If I had no matches, then throw an error

Claywd (approved):
looks great. Could be optimized a smidgen. I find it helpful to provide unique error messages for each err handling statement.

internal/common/common.go
fharper: Good call!

internal/common/common.go
```diff
@@ -33,17 +49,44 @@ func CheckForVersionUpdate() {
...
+			latestVersion = matches[1]
+			fmt.Printf(latestVersion)
+		} else {
+			fmt.Printf("checking for a newer version failed (no reference to kubefirst release) with: %s", err)
+			return nil, true
+		}
+	} else {
+		fmt.Printf("checking for a newer version failed (HTTP error) with: %s", err)
 		return nil, true
 	}
```
Claywd, line 84: What you have is plenty good. Here is another way to write it such that you handle the errors and then proceed executing code without having to wrap it in and if condition. Faster in some cases but also more readable and flat. 
```suggestion
if resp.StatusCode != http.StatusOK {
		fmt.Printf("checking for a newer version failed (HTTP error) with: %s", err)
		return nil, true
	}

	bodyBytes, err := io.ReadAll(resp.Body)
	if err != nil {
		fmt.Printf("checking for a newer version failed (cannot read the file) with: %s", err)
		return nil, true
	}

	bodyString := string(bodyBytes)

	if !strings.Contains(bodyString, "url \"https://github.com/kubefirst/kubefirst/archive/refs/tags/") {
		fmt.Printf("checking for a newer version failed (no reference to kubefirst release) with: %s", err)
		return nil, true
	}
	
	re := regexp.MustCompile(`.*/v(.*).tar.gz"`)
	matches := re.FindStringSubmatch(bodyString)
	latestVersion = matches[1]
	fmt.Printf(latestVersion)
```

Claywd (approved):
Looks good. One piece of advice. Both ways work in the same time complexity. Feel free to merge what you have!

internal/common/common.go
fharper: Is it a best practice in Go? I'm like a junior Go developer, it's fun 😄

In other languages, I wouldn't do that many return, nor would isolate the error handling.

PR #1768, opened 2023-08-25
fix: use force-destroy rather than force_destroy

cameronraysmith:
feel free to close without merge


cameronraysmith:
Resolved in https://github.com/kubefirst/kubefirst/commit/b16faa5a8f9ef83cc08eef271bd8b0370d4b1e80 .

PR #1815, opened 2023-09-15
fix: don't open the browser with console with the ci flag (fix #1713)

### LINKED ISSUE #1713: Suppress "helpful box" on CLI creates for easier scripting
### What is your feature idea?

Can you provide a `quiet` type flag for running a CLI create so that the helpful box at the end does not appear?

### Why is it needed?

This box really is helpful...to humans. But in the world of experienced kubefirst users and/or scripting it requires exiting before continuing on with the workflow.  Ideally I could provide a flag to the create to say "please just create and then give me back to the terminal!".

### Is this missing feature preventing you from using kubefirst?

- [ ] Yes

### Code of Conduct

fharper:


fharper:
I'll close my PR, as the next release will have a refactor CLI, which won't open by default the console in the browser, and won't make the process wait for the user to terminate the handoff screen.

PR #1850, opened 2023-10-22
refactor: change vars requirement validation

jairoFernandez:


fharper:
Confirmed by @jairoFernandez that it's not relevant anymore.

PR #1873, opened 2023-10-27
chore: remove other installation instructions as moved to the docs

fharper:


fharper:
To be merged only when https://github.com/kubefirst/docs/pull/559 is.

johndietz (approved):

PR #1924, opened 2023-11-23
chore: add a question about gitops-template being a fork or not (#1919)

fharper:


fharper:
Don't think I can rebase main with a PR, and I don't have admin access, so we'll keep the messy logs from my last issue template fixes

PR #1944, opened 2023-12-07
feat: remove the ability to run cluster zero on atlas

johndietz:
it doesn't make sense to keep cluster zero data in more permanent storage. we don't want to keep the data and should not encourage its longevity.

### DIFF
--- a/internal/launch/constants.go
+++ b/internal/launch/constants.go
@@ -11,7 +11,7 @@ const (
 	helmChartName     = "kubefirst"
 	helmChartRepoName = "kubefirst"
 	helmChartRepoURL  = "https://charts.kubefirst.com"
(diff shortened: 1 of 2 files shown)


internal/launch/constants.go
```diff
@@ -11,7 +11,7 @@ const (
 	helmChartName     = "kubefirst"
 	helmChartRepoName = "kubefirst"
 	helmChartRepoURL  = "https://charts.kubefirst.com"
-	helmChartVersion  = "2.3.6-rc24"
+	helmChartVersion  = "2.3.6-rc45"
```
johndietz, line 14:
```suggestion
	helmChartVersion  = "2.3.6-rc51"
```

jarededwards (approved):

Issue #2176, opened 2024-05-12
Running kubefirst changes my stty settings


mwoodpatrick:
### Which version of kubefirst are you using?

kubefirst-cli golang utility version: v2.4.6

### Which cloud provider?

k3d (local)

### Which DNS?

None specific

### Which installation type?

CLI

### Which distributed Git provider?

GitHub

### Did you use a fork of `gitops-template`?

No

### Which Operating System?

None specific

### What is the issue?

Running the following kubefirst command:

`kubefirst k3d create --cluster-name kubefirst_test1 --github-user mwoodpatrick --git-provider github
`

changes my stty settings so nothing is displayed to my terminal until I run:

`tset`

Changes seen in stty settings (stty -a)

```
 diff stty.kubefirst stty.orig
6c6
< -ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr -icrnl -ixon -ixoff
---
> -ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl ixon -ixoff
9c9
< -isig -icanon -iexten -echo echoe echok -echonl -noflsh -xcase -tostop -echoprt
---
> isig icanon iexten echo echoe echok -echonl -noflsh -xcase -tostop -echoprt
```

Not all kubefirst commands do this e.g.:

` kubefirst version`
 
  

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

fharper:
@mwoodpatrick: thanks for reporting this issue. I can replicate this, but I have no idea why we would modify stty settings, so let me check with the engineering team.

mwoodpatrick:
Any updates on this issue?

fharper:
@fharper no sorry, the team is quite swamped at the moment, and didn't have time to investigate to find what is changing stty.

CristhianF7:
I was not able to reproduce the issue, I don't see any stty difference even the k3d commands returns the same error `exit status 1`

![Screenshot 2024-05-29 at 6 23 03 PM](https://github.com/kubefirst/kubefirst/assets/14883029/783fe627-f0d7-4103-8d93-dba04742ac5b)

Here's the stty diff before and after running `k3d create` 

```bash
1c1
[...]

PR #2220, opened 2024-07-02
added bubbletea quit method instead of os exit

jokestax:
in stty there is echo option which is set off when bubbletea is working,so if we do bubbletea.Quit() it sets it to on and then quits

### DIFF
--- a/cmd/k3d/create.go
+++ b/cmd/k3d/create.go
@@ -31,6 +31,7 @@ import (
 	"github.com/kubefirst/kubefirst/internal/catalog"
 	"github.com/kubefirst/kubefirst/internal/gitShim"
 	"github.com/kubefirst/kubefirst/internal/segment"
+	"github.com/kubefirst/kubefirst/internal/progress"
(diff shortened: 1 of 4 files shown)


fharper:
I'll test it.

A quick note on the PR:
- When squashing, add (fix #2176) to the commit title so it close the issue at the same time please.
- Also, for the title, add the `fix: ` prefix to it. I need to update CONTRIBUTING.md to write about this, but we use the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).

fharper:
LGTM

fharper (approved):
You tested it, @CristhianF7 also, and I, and it fixed the issue so from my POV, we're good to merge this. If the reporter for the initial issue see any problems we can revisit, but as far as I'm concerned, this issue is fixed, thanks you both, it was such an annoying bug, worth releasing if you ask me :)

CristhianF7 (approved):

PR #2248, opened 2024-07-18
feat : custom repositories

jokestax:
we can now deploy multiple repos on same organistation using below  flags : 

 "--gitopsRepoName" metaphorRepoName "adminTeamName" "developerTeamName"

but we have to use kubefirst-api repo branch "feat-custom-repo" and kubefirst-gitops-template repo branch "feat-custom-repo" until its merged into main


mrsimonemms:
Change looks good, but a couple of nits:
* can you run `go fmt ./...` as the formatting is all out on the changes (this may change additional files which shouldn't be updated here - just the lines you've changed)
* there's a merge conflict which needs resolving

fharper:
@jokestax can you fix the conflict before we review this please?

jokestax:
Fixed it in the recent push

fharper:
We won't be able to merge this PR, you need to fix the conflict at the Git level.

CleanShot 2024-07-23 at 11 32 51@2x

jokestax:
@fharper, done

fharper:
@jokestax you wrote done, but it's a draft, just want to be sure it's ready for review? :)

fharper:
oh, should I wait for k3d support?

jokestax:
Now it's supported :)

fharper:
Nice, undraft this PR when you are ready for us to review it :)

jokestax:
its ready 👍

fharper:
@jokestax perfect, I'll try to test it tomorrow.

fharper:
@jokestax: can you check why CodeQL is failing also. All those should be green before anyone take the time to review the code changes or test :)

jokestax:
No Fred,this fix is dependent on changes done in api too,so when anyone who is testing they need to change their go mod file by "replacing Kubefirst api v0.1.26 to local Kubefirst api which is on branch feat-custom-repo"

fharper:
Perfect, we will also need to reflect these changes in the console UI so we don't add more disparities between the CLI and the UI. I'm testing those changes now.

fharper:
The docs also needs to be updated to reflect these new flags.

jokestax:
Yes Fred, for frontend I will coordinate with Cris

fharper:
So I did the first test with k3d & GitHub, and I got this error:

CleanShot 2024-07-25 at 10 53 26@2x

I'm running the CLI locally on this branch, with the API locally on your branch also.

Also, may I suggest we follow the flag name convention we seems to be using, which is not camel case, but kebab case.

fharper:
I got the same error with Civo, so maybe I'm doing something wrong. How did you test this PR @jokestax ?

jokestax:
clone kubefirst-api too ,checkout branch "feat-custom-repo" and in the go mod file of kubefirst add this line "replace github.com/kubefirst/kubefirst-api v0.1.26 =>  <PATH-TO-KUBEFIRST-API>" ,replace  <PATH-TO-KUBEFIRST-API> with actual path

fharper:
Yes, as I wrote I used the API branch also. @CristhianF7 told me last time that we cannot replace the API in mod file to test locally anymore, just running the API locally do the trick (this is how I successfully tested all previous PR recently).

cmd/akamai/command.go
```diff
@@ -31,6 +31,10 @@ var (
 	gitProtocolFlag          string
 	gitopsTemplateURLFlag    string
 	gitopsTemplateBranchFlag string
+	gitopsRepoName			 string
```
mrsimonemms, line 34: Can you run the `go fmt ./...` command so that the formatting is correctly applied to this change please? You only need to change files/lines you've updated.

fharper:
So I tried with k3d again with the `replace` in `go.mod` and it moved passed the issue I had, but I got another one.

CleanShot 2024-07-25 at 13 44 45@2x

A Terraform error that seems to check if the default repository exists, not the custom names I choose.

CleanShot 2024-07-25 at 13 44 59@2x

I don't have `mygitops` or `mymetaphor` repositories on my account, but I have the default ones.

CleanShot 2024-07-25 at 13 47 19@2x

So I wonder if there is a check that doesn't take into account the name passed with the new flags. I can replicate by just having the `gitops` & `metaphor` repositories name used on my account. If I delete them, this error doesn't happen, but it create the repositories with the default name, and give me this error (which mean that I'm using your branches).

CleanShot 2024-07-25 at 13 51 21@2x

I'll test with a public cloud since the behavior may be different.

jokestax:
You need to use th flag "--gitops-template-branch feat-custom-repo" too :-;

fharper:
> --gitops-template-branch feat-custom-repo

Oh, I missed that, my bad 🤦‍♂️

jokestax:
And for clouds you need to run this locally

fharper:
I suggest this PR also modify this error message by mentioning the new flags as a second possibility when there is this error.

CleanShot 2024-07-25 at 17 30 02@2x

fharper:
I'm doing other tests today in-between other tasks (sorry, I have multiple things to finish before next week, and others can also test this PR, not just me), but there are still two discussions open that will need to be fix (I tag you on those @jokestax ).

fharper:
Also, I tested k3d with the latest changed, and those are the repositories that were created, so it's not working anymore.

CleanShot 2024-07-29 at 10 31 03@2x

fharper:
I'll wait for this to be fixed before testing other clouds as I assume the changes were similar in API for non-k3d. Let me know when I can give it another go please.

jokestax:
@fharper can you comment the command you ran?

fharper:
CleanShot 2024-07-29 at 10 37 49@2x

Even if the command was wrong, the end result shouldn't be those repository though.

jokestax:
I think I know where this issue is ,will fix it by tonight Can you try with Civo once ?

fharper:
I'll wait for you to resolve the two things pending before doing another test:
- The code formatting @mrsimonemms mentioned a while ago.
- The flag naming convention I mentioned a couple of days ago.

fharper:
Also, better add a new comment when you to update the situation instead of updating a previous one as nobody will get notification on an updated comment :)

jokestax:
> I suggest this PR also modify this error message by mentioning the new flags as a second possibility when there is this error.
> 
> CleanShot 2024-07-25 at 17 30 02@2x

@fharper ,you want the alert of message of flag feature here?

internal/common/common.go
```diff
@@ -113,17 +114,68 @@ func GetRootCredentials(cmd *cobra.Command, args []string) error {
...
+	
+
+	if !gitopsFound {
+		log.Info().Msg("Gitops Repo not found")
+		return "cantfindgit","cantfindmeta"
+	}
+
+	if !metaphorFound {
+		log.Info().Msg("Metaphor Repo not found")
+		os.Exit(1)
```
patrickdappollonio, line 160: Happy to chat about this one, but this seems wrong: exiting right here means that several pieces of the Go code here won't be executed (like `defer` or similar).

It's better to return errors that the callers can decide whether or not they need to make an action out of it. As a golden rule, the only place that should exit with an exit code is the top-most `main()` function from the Go application. As a package (called `common` here) it's better to return the error so the caller handles it and decides if it's so critical that the app needs to exit and crash.

tools/aws-assume-role.sh
```diff
@@ -19,7 +19,7 @@
 #
 # Change the AWS account ID & role name
 #
-ROLE="arn:aws:iam::111111111111:role/AssumedAdmin"
+ROLE="arn:aws:sts::126827061464:assumed-role/KubernetesAdmin/rishi-wins"
```
patrickdappollonio, line 22: This might've been an unexpected change. I would revert this to what it was before.

patrickdappollonio (changes requested):
Some notes below. Happy to chat about any of them!

I also feel the sentiment from Simon, this PR definitely needs a `gofmt` pass.

fharper:
> @fharper ,you want the alert message of flag feature here?

This is just a suggestion, but before, when one had no choice on the name, they either had to delete them or choose another org/account, but now they can rename it, so the error could be something along the lines of "Error: the following repositories already exist: you must delete them before continuing with your kubefirst installation or choose another name for them using the [INSERT FLAGS HERE]"

internal/common/common.go
```diff
@@ -113,17 +114,68 @@ func GetRootCredentials(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func Getgitmeta(clusterName string) (gitopsRepoName string, metaphorRepoName string) {
+	var gitopsFound, metaphorFound bool
+	homePath,err := os.UserHomeDir()
```
patrickdappollonio, line 119: Just realized this error is being shadowed/ignored.

fharper:
@jokestax: before I test, can you resolve the conflicts of this branch please?

Also, can you rebase and sign the commits that aren't. We will squash this PR at the end as it doesn't make sense to merge all this history, but it will prevent the action to complain, and anyway, it's just good practice :)

fharper:
@jokestax: also be sure to resolve the discussion or suggestions in this PR as we won't be able to merge it without this (branch protection), but also because it ensure the author that those were fixed :)

fharper (changes requested):
- Resolve conversations.
- Sign unsigned commits.
- Fix the merge conflict.

patrickdappollonio (changes requested):

cmd/k3d/create.go
```diff
@@ -1038,7 +1069,28 @@ func runK3d(cmd *cobra.Command, args []string) error {
...
+					return err
+				}
+				log.Info().Msgf("Error creating Argo CD application on attempt number #%d: %v\n", attempt, err)
+				time.Sleep(5 * time.Second)
+				continue
+			}
+
+			log.Info().Msgf("Argo CD application created successfully on attempt #%d: %s\n", attempt, app.Name)
+			break
+		}
```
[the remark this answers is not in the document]
jokestax: sure patrick will use that,this code i had pushed long like last week,will make those changes

mrsimonemms:
Can this be closed?

PR #2269, opened 2024-08-28
fix:add restart function for argo

jokestax:
## Description
added restart function for argo controller,which sometimes starts before argo server and fails in waiting secret

## Related Issue(s)
this resolves timeout issues

## How to test
build the binary and run ./kubefirst k3d create


cmd/k3d/create.go
```diff
@@ -1034,12 +1035,25 @@ func runK3d(cmd *cobra.Command, args []string) error {
...
+			return true, nil
+		})
 
 		if err != nil {
 			return fmt.Errorf("error creating argocd application : %w", err)
 		}
 
+		log.Info().Msg("Argo CD application created successfully\n")
+
 		log.Info().Msg("Argo CD application created successfully\n")
```
patrickdappollonio, line 1057: These two lines seem duplicated. Also, "log" statements don't need linebreaks.

cmd/k3d/create.go
```diff
@@ -1034,12 +1035,25 @@ func runK3d(cmd *cobra.Command, args []string) error {
...
+			return fmt.Errorf("error in restarting argocd controller %w", err)
+		}
+
+		err = wait.PollImmediate(5*time.Second, 20*time.Second, func() (bool, error) {
+			_, err := argocdClient.ArgoprojV1alpha1().Applications("argocd").Create(context.Background(), registryApplicationObject, metav1.CreateOptions{})
+			if err != nil {
+				return false, nil
+			}
+			return true, nil
+		})
```
patrickdappollonio, line 1049: What's the expectation here? If the application couldn't be created, try again?

If so, can we scope down the rules a bit? If it couldn't be created cause it already exists, for example, would have no point on keep trying, right?

If you can give some more context I can scope it down.

patrickdappollonio (changes requested):

patrickdappollonio (approved):

PR #2271, opened 2024-08-28
use generated release notes for gitops-template

jarededwards:


.github/workflows/releaser.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Update version file
         run: echo $GITHUB_REF_NAME > VERSION.md
       - name: Release kubefirst/gitops-template
```
jarededwards, line 19:
```suggestion
      - name: Release konstructio/gitops-template
```

PR #2279, opened 2024-09-03
Clean up code using GolangCI-Lint

patrickdappollonio:
## Related Issue(s)

Fixes #


internal/gitShim/init.go
```diff
@@ -75,45 +74,42 @@ func InitializeGitProvider(p *GitInitParameters) error {
...
 
-		// Check for existing base projects
-		// Save for detokenize
 		subgroups, err := gitlabClient.GetSubGroups()
 		if err != nil {
-			log.Fatal().Msgf("couldn't get gitlab subgroups for group %s: %s", p.GitOwner, err)
+			return fmt.Errorf("couldn't get GitLab subgroups for group %q: %w", p.GitOwner, err)
 		}
-		for _, teamName := range p.Repositories {
+		for _, teamName := range p.Teams {
```
patrickdappollonio, line 109: When passing it to ChatGPT, it changed this from `Repositories` to `Teams`. I think GPT is right, no? This should iterate over teams?

@jarededwards @CristhianF7

patrickdappollonio: Seems it was an incorrect statement so the code was updated to reflect the changed value.

jarededwards (approved):

PR #2284, opened 2024-09-16
fix reset cmd

### LINKED ISSUE #2282: kubefirst reset fails when .kubefirst file is not present
### Which version of kubefirst are you using?

v2.6.0

### Which cloud provider?

None specific

### Which DNS?

None specific

### Which installation type?

None specific

### Which distributed Git provider?

None specific

### Did you use a fork of `gitops-template`?

No

### Which Operating System?

macOS

### What is the issue?

kubefirst reset should guard against directories and files not being present:

```
➜  ~ kubefirst version  
                                                                              
                                                                              
  kubefirst-cli golang utility version: v2.6.0                                


➜  ~ kubefirst reset  
Error: unable to determine contents of kubefirst-checks: unexpected type <nil>

If a detailed error message was available, please make the necessary corrections before retrying.
You can re-run the last command to try the operation again.

➜  ~ rm -rf ~/.k1 ~/.kubefirst
➜  ~ kubefirst reset          
Error: unable to determine contents of kubefirst-checks: unexpected type <nil>

If a detailed error message was available, please make the necessary corrections before retrying.
You can re-run the last command to try the operation again.


```

### Code of Conduct

jokestax:
## Description
Handle the nil value of viperfile content

## Related Issue(s)

Fixes #2282 

## How to test
build the binary and run "./kubefirst reset"

### DIFF
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -12,7 +12,6 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/charmbracelet/glamour"
 	"github.com/konstructio/kubefirst-api/pkg/progressPrinter"
 	utils "github.com/konstructio/kubefirst-api/pkg/utils"
 	"github.com/konstructio/kubefirst/internal/progress"
@@ -34,10 +33,8 @@ var resetCmd = &cobra.Command{
 		switch v := checksMap.(type) {
 		case nil:
 			// Handle the nil case explicitly
-			in := `# Successfully reset`
-			out, _ := glamour.Render(in, "dark")
-			fmt.Print(out)
-			progress.Progress.Quit()
+			message := `# Succesfully reset`
+			progress.Success(message)
 			return nil
 		case string:
 			if v == "" {
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/charmbracelet/glamour"
 	"github.com/konstructio/kubefirst-api/pkg/progressPrinter"
 	utils "github.com/konstructio/kubefirst-api/pkg/utils"
 	"github.com/konstructio/kubefirst/internal/progress"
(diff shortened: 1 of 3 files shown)


johndietz (dismissed):

cmd/reset.go
```diff
@@ -30,6 +31,13 @@ var resetCmd = &cobra.Command{
 		cloudProvider := viper.GetString("kubefirst.cloud-provider")
 
 		checksMap := viper.Get("kubefirst-checks")
+		if checksMap == nil {
+			in := `# Succesfully reset`
+			out, _ := glamour.Render(in, "dark")
+			fmt.Print(out)
+			progress.Progress.Quit()
+			return nil
+		}
```
patrickdappollonio, line 40: Down below there seems to be a switch on the type, which could have the type to be empty (`nil`).

Do we need to quit if we can't find it? Or perhaps we need to ignore it if it's not there but continue with the cleanup after?

patrickdappollonio (changes requested):

cmd/reset.go
patrickdappollonio: Strange, because right down below, it says "if the map is a string, and that string is empty, log a message but continue".

What makes the map being nil vs being empty different?

cmd/reset.go
```diff
@@ -30,6 +31,13 @@ var resetCmd = &cobra.Command{
 		cloudProvider := viper.GetString("kubefirst.cloud-provider")
 
 		checksMap := viper.Get("kubefirst-checks")
+		if checksMap == nil {
+			in := `# Succesfully reset`
+			out, _ := glamour.Render(in, "dark")
+			fmt.Print(out)
+			progress.Progress.Quit()
+			return nil
```
CristhianF7: Please use the progress success function

```suggestion
		message := `# Succesfully reset`
		progress.Success(message)
```

CristhianF7 (changes requested):

cmd/reset.go
jokestax: if we have .kubefirst file present,then in that file we will have "kubefirst-checks" as one of the fields,which will be set to empty as "kubefirst-checks:""' and we run a reset cmd,however when we dont have that file present at all go thows nil error

jokestax: we can remove the else in case string ig

patrickdappollonio (approved):

CristhianF7 (approved):

PR #2311, opened 2024-10-23
fix: retry

CristhianF7:
## Description

- Handles secret creation error properly `secrets "container-registry-auth" already exists`
- Added execution control for post-detokenize process

### DIFF
--- a/cmd/k3d/create.go
+++ b/cmd/k3d/create.go
@@ -1187,34 +1187,40 @@ func runK3d(cmd *cobra.Command, _ []string) error {
 	progressPrinter.AddTracker("wrapping-up", "Wrapping up", 2)
 	progressPrinter.SetupProgress(progressPrinter.TotalOfTrackers(), false)
 
-	if err := k3d.PostRunPrepareGitopsRepository(config.GitopsDir); err != nil {
-		return fmt.Errorf("error detokenizing post run: %w", err)
-	}
+	executionControl = viper.GetBool("kubefirst-checks.post-detokenize")
+	if !executionControl {
 
-	gitopsRepo, err := git.PlainOpen(config.GitopsDir)
-	if err != nil {
-		return fmt.Errorf("error opening repo at %q: %w", config.GitopsDir, err)
-	}
-	_, err = os.Stat(fmt.Sprintf("%s/terraform/%s/remote-backend.md", config.GitopsDir, config.GitProvider))
-	if err == nil {
-		err = os.Rename(fmt.Sprintf("%s/terraform/%s/remote-backend.md", config.GitopsDir, config.GitProvider), fmt.Sprintf("%s/terraform/%s/remote-backend.tf", config.GitopsDir, config.GitProvider))
-		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/k3d/create.go
```diff
@@ -1187,34 +1187,39 @@ func runK3d(cmd *cobra.Command, _ []string) error {
...
-			return fmt.Errorf("failed to rename remote-backend.md to remote-backend.tf: %w", err)
+			return fmt.Errorf("error opening repo at %q: %w", config.GitopsDir, err)
+		}
+		_, err = os.Stat(fmt.Sprintf("%s/terraform/%s/remote-backend.md", config.GitopsDir, config.GitProvider))
+		if err == nil {
+			err = os.Rename(fmt.Sprintf("%s/terraform/%s/remote-backend.md", config.GitopsDir, config.GitProvider), fmt.Sprintf("%s/terraform/%s/remote-backend.tf", config.GitopsDir, config.GitProvider))
+			if err != nil {
+				return fmt.Errorf("failed to rename remote-backend.md to remote-backend.tf: %w", err)
+			}
 		}
```
patrickdappollonio, line 1206: This seems like the opposite of what you want to do and it's considered an anti-pattern in Go: not only it eats the error, it also handles the non-error like an error.

If you can't stat the file, do you still continue?

internal/gitShim/containerRegistryAuth.go
```diff
@@ -48,12 +49,20 @@ func CreateContainerRegistrySecret(obj *ContainerRegistryAuth) (string, error) {
...
 		}
-		if err := k8s.CreateSecretV2(obj.Clientset, argoDeployTokenSecret); err != nil {
+		err := k8s.CreateSecretV2(obj.Clientset, argoDeployTokenSecret)
+		if errors.IsAlreadyExists(err) {
+			if err := k8s.UpdateSecretV2(obj.Clientset, "argo", secretName, data); err != nil {
+				return "", fmt.Errorf("error while updating secret for GitHub container registry auth: %w", err)
+			}
+		}
+
+		if err != nil {
```
patrickdappollonio, line 65: You might wanna check too for...

```suggestion
		if err != nil && !errors.IsAlreadyExists(err) {
```

The `errors.IsAlreadyExists()` has a tree: if it already exists and you updated the flow successfully, `err` will still be true to `errors.IsAlreadyExists()` when you hit this line. The `err` on `UpdateSecretV2` is scoped to the `if`.

cmd/k3d/create.go
CristhianF7: I didn't write this code, I only wrapped it in a condition so it doesn't run again for retries. Other than that I Agree

internal/gitShim/containerRegistryAuth.go
CristhianF7: Yup, Good catch. for some reason I removed that last night and don't recall why

cmd/k3d/create.go
CristhianF7: I'll remove the Stat since we have talked a lot about this 😄

patrickdappollonio (approved):

jokestax (approved):

PR #2317, opened 2024-11-04
fix: delete k1dir instead of kubeconfig file

jokestax:
## Description
when we do "kubefirst k3d destroy",it doesnt delete the k1dir repo which return an error that "repo already exists" when we try to provision a k3d cluster again and the cleanup path is also wrong one, it should be (config.K1Dir + "<cluster_name>/kubeconfig")

## Testing
Provision a Kubefirst in k3d environment
Clone the repo
run:
```bash
go build .
./kubefirst k3d destroy
```

### DIFF
--- a/cmd/k3d/destroy.go
+++ b/cmd/k3d/destroy.go
@@ -223,9 +223,9 @@ func destroyK3d(_ *cobra.Command, _ []string) error {
 		viper.WriteConfig()
 	}
 
-	if _, err := os.Stat(config.K1Dir + "/kubeconfig"); !os.IsNotExist(err) {
-		if err := os.Remove(config.K1Dir + "/kubeconfig"); err != nil {
-			return fmt.Errorf("unable to delete %q: %w", config.K1Dir+"/kubeconfig", err)
+	if _, err := os.Stat(config.K1Dir); !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


cmd/k3d/destroy.go
```diff
@@ -223,9 +223,9 @@ func destroyK3d(_ *cobra.Command, _ []string) error {
 		viper.WriteConfig()
 	}
 
-	if _, err := os.Stat(config.K1Dir + "/kubeconfig"); !os.IsNotExist(err) {
-		if err := os.Remove(config.K1Dir + "/kubeconfig"); err != nil {
-			return fmt.Errorf("unable to delete %q: %w", config.K1Dir+"/kubeconfig", err)
+	if _, err := os.Stat(config.K1Dir); !os.IsNotExist(err) {
```
johndietz, line 226: i understand the desire to remove the k1dir. i want to also question that this is a change instead of an addition. why don't we want to remove the kubeconfig any longer?

jokestax: the ultimate goal to make a k3d provisioning possible again is to to remove the directory,i am not sure what was the thought behind removing kubeconfig,but this change should work smoothly for k3d reprovisioning

mrsimonemms (approved):

PR #2321, opened 2024-11-11
feat: disable bubbletea

jokestax:
## Description
add ci flag so to test the provisioning of cluster in argo workflows 

## How to test
provision any cluster using cli and add additional flag "--ci"

### DIFF
--- a/internal/progress/ci.go
+++ b/internal/progress/ci.go
@@ -1,0 +1,41 @@
+package progress
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/konstructio/kubefirst/internal/cluster"
+)
+
+func WatchClusterForCi(clusterName string) {
+
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+
+	done := make(chan bool)
+
+	go func() {
+		for {
+			select {
+			case <-done:
+				return
+			case <-ticker.C:
+				provisioningCluster, _ := cluster.GetCluster(clusterName)
+
+				if provisioningCluster.Status == "error" {
+					log.Fatalf("unable to provision cluster: %s", provisioningCluster.LastCondition)
+					done <- true
+				}
+
+				if provisioningCluster.Status == "provisioned" {
+					fmt.Println("\n cluster has been provisioned via ci")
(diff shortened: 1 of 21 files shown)


internal/progress/ci.go
```diff
@@ -0,0 +1,39 @@
...
+				if provisioningCluster.Status == "provisioned" {
+					fmt.Println("\n cluster has been provisioned via ci")
+					fmt.Printf("\n kubefirst URL: https://kubefirst.%s \n", provisioningCluster.DomainName)
+					done <- true
+				}
+			}
+		}
+	}()
+
+	<-done
```
patrickdappollonio, line 38: Ok this is a bit of a nasty cheat that more than one might have a hard time using in the future 😆

You're waiting here because nobody has sent data to it. But you also have the read inside the goroutine inside the case statement! Which one wins when `done <- true` happens inside the `ticker.C`?

The short answer is the asynchronous one, at the top level that simply returns. The second one, the one inside the `for select` will need an extra CPU cycle to leave the statement of the `ticker.C` then go back to the parent one, to finally read the "return" you have there, but at that point, the function is long gone.

May I propose you to use an `errgroup` or the tool we've been using so far for waiting for events from happening instead of generating yet another function "that waits for a condition"?

The other part this doesn't cover is that there is no timeout. This function will run forever if none of the conditions break the loop.

main.go
```diff
@@ -24,8 +24,8 @@ import (
 
 func main() {
 	argsWithProg := os.Args
-
-	bubbleTeaBlacklist := []string{"completion", "help", "--help", "-h", "quota", "logs"}
+	fmt.Println(argsWithProg)
```
patrickdappollonio, line 27: This seems like a debugging statement, might want to remove it.

patrickdappollonio (changes requested):
Adding to the comments I left, it seems possible to disable app-wide bubbletea from being "interactive" by setting the `WithoutRenderer()` flag. That imho is better than sprinkling in 300 places the `if canRunBubbleTea` statement:

https://github.com/charmbracelet/bubbletea/blob/1feb60b44b74d9a3a7dc54b90ffbecc8ffd6b40d/options.go#L169-L181

PR #2322, opened 2024-11-12
fix: bubbletea output

CristhianF7:
## Description
- Adds `tea.WithoutRenderer()`for CI executions.

### DIFF
--- a/cmd/akamai/command.go
+++ b/cmd/akamai/command.go
@@ -18,7 +18,6 @@ import (
 var (
 	// Create
 	alertsEmailFlag          string
-	ciFlag                   bool
 	cloudRegionFlag          string
 	clusterNameFlag          string
 	clusterTypeFlag          string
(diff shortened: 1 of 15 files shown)


cmd/akamai/command.go
```diff
@@ -128,6 +128,7 @@ func RootCredentials() *cobra.Command {
 	authCmd.Flags().BoolVar(&copyArgoCDPasswordToClipboardFlag, "argocd", false, "copy the ArgoCD password to the clipboard (optional)")
 	authCmd.Flags().BoolVar(&copyKbotPasswordToClipboardFlag, "kbot", false, "copy the kbot password to the clipboard (optional)")
 	authCmd.Flags().BoolVar(&copyVaultPasswordToClipboardFlag, "vault", false, "copy the vault password to the clipboard (optional)")
+	authCmd.Flags().BoolVar(&ciFlag, "ci", false, "if running kubefirst in ci, set this flag to disable interactive features")
```
patrickdappollonio, line 131: PSA: you can create this flag on the root command as a `PersistentFlag` and it will be inherited by every subcommand. That way you don't have to remember to maintain it everywhere and just on a single place.

PR #2328, opened 2024-11-25
fix: delete k1dir instead of kubeconfig file

jokestax:
## Description
when we do "kubefirst reset",it doesnt delete the k1dir repo which return an error that "repo already exists" when we try to provision a cluster with same name again,this will fix it

### DIFF
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -127,9 +127,9 @@ func runReset() error {
 		return fmt.Errorf("error writing viper config: %w", err)
 	}
 
-	if _, err := os.Stat(k1Dir + "/kubeconfig"); !os.IsNotExist(err) {
-		if err := os.Remove(k1Dir + "/kubeconfig"); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/reset.go
```diff
@@ -127,9 +127,9 @@ func runReset() error {
 		return fmt.Errorf("error writing viper config: %w", err)
 	}
 
-	if _, err := os.Stat(k1Dir + "/kubeconfig"); !os.IsNotExist(err) {
-		if err := os.Remove(k1Dir + "/kubeconfig"); err != nil {
-			return fmt.Errorf("unable to delete %q folder, error: %w", k1Dir+"/kubeconfig", err)
+	if _, err := os.Stat(k1Dir); !os.IsNotExist(err) {
```
patrickdappollonio, line 130: No need to stat the file if you're going to delete it with `os.RemoveAll()`.

patrickdappollonio (changes requested):

patrickdappollonio (approved):

PR #2350, opened 2024-12-20
[WIP] chore: restructure CLI init and begin logging refactor

nathan-nicholson:
## Description
Very WIP stab at refactoring the foundation of how dependencies such as logging and message printing are handled by commands.

## Related Issue(s)

Fixes #


internal/cluster/cluster_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
+	apiTypes "github.com/konstructio/kubefirst-api/pkg/types"
+)
+
+type mockHTTPClient struct {
```
patrickdappollonio, line 13: Not something that needs change but don't forget there's a whole `httptest` package that adds support for mocking stuff out of the box (including http requests and responses). 

See here for reference: https://github.com/konstructio/dropkick/blob/main/internal/civo/sdk/json/json_fetch_test.go

nathan-nicholson: Good call. This was from the robot and I haven't got back to this yet.

PR #2352, opened 2024-12-30
feat: add bottlerocket ami types

jokestax:
## Description

Add support for **Bottlerocket** AMI types in the AWS cloud CLI.

## Testing

Provision a cluster using the CLI with the `--ami-type <AMI_TYPE>` flag. 

### Example Command:
```bash
go run . aws create \
  --alerts-email 1@gmail.com \
  --domain-name kuberfu.com \
  --dns-provider cloudflare \
  --gitops-template-branch main \
  --git-provider github \
  --github-org ar6464 \
  --cluster-name rr18 \
  --ami-type BOTTLEROCKET_ARM_64 
```

## Supported AMI Types

| AMI Type                     |
|------------------------------|
| `AL2_x86_64`                 |
| `AL2_ARM_64`                 |
| `BOTTLEROCKET_ARM_64`        |
| `BOTTLEROCKET_x86_64`        |
| `BOTTLEROCKET_ARM_64_NVIDIA` |
| `BOTTLEROCKET_x86_64_NVIDIA` |


cmd/aws/create.go
```diff
@@ -161,7 +156,125 @@ func ValidateProvidedFlags(gitProvider string) error {
...
+	output, err := ssmClient.GetParameter(context.TODO(), input)
+	if err != nil {
+		return "", fmt.Errorf("failed to initialise ssm client: %w", err)
+	}
+
+	return *output.Parameter.Value, nil
+}
+
+func GetAMIArchitecture(cfg aws.Config, amiID string) (string, error) {
+	ec2Client := ec2.NewFromConfig(cfg)
```
nathan-nicholson, line 242: Similar to other clients, this should be passed in as a dependency and not instantiated in-line. Makes it basically impossible to test or for anyone else to mock it if necessary.

cmd/aws/create.go
```diff
@@ -161,7 +156,125 @@ func ValidateProvidedFlags(gitProvider string) error {
...
+	if len(output.Images) == 0 {
+		return "", fmt.Errorf("no images found for AMI ID: %s", amiID)
+	}
+
+	val := output.Images[0]
+	return string(val.Architecture), nil
+}
+
+func GetSupportedInstanceTypes(cfg aws.Config, architecture string) ([]string, error) {
+	ec2Client := ec2.NewFromConfig(cfg)
```
nathan-nicholson, line 260: Same as other ec2 client comment. Since we seem to be doing several sequential interactions with the ec2Client, it probably makes more sense to have these as methods to a struct responsible for interacting with ec2.

internal/utilities/flags.go
```diff
@@ -131,6 +131,13 @@ func GetFlags(cmd *cobra.Command, cloudProvider string) (types.CliFlags, error)
 		}
 
 		cliFlags.ECR = ecrFlag
+
+		amiType, err := cmd.Flags().GetString("ami-type")
+		if err != nil {
+			progress.Error(err.Error())
```
nathan-nicholson, line 137: We're in the process of ripping out bubbletea, so lets not add more.

cmd/aws/create.go
```diff
@@ -161,7 +156,125 @@ func ValidateProvidedFlags(gitProvider string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get AMI architecture: %w", err)
+	}
+
+	instanceTypes, err := GetSupportedInstanceTypes(cfg, architecture)
+	if err != nil {
+		return fmt.Errorf("failed to get supported instance types: %w", err)
+	}
+
+	for _, instanceType := range instanceTypes {
```
nathan-nicholson, line 220: Nit: Since we're dealing with strings and string slices, we should be able to use the slices module in the std lib: https://cs.opensource.google/go/go/+/go1.23.4:src/slices/slices.go;l=117

`Contains` should do the job. This would also allow for us to use a "not contains" with an error and return nil at the end.

nathan-nicholson (changes requested):
There are a lot of functions with in-line defined dependencies (primarily clients to AWS) that will make this almost impossible to properly test logic paths without a lot of effort and real resources.

I may take a stab at refactoring/pairing on this, but I'm not comfortable with merging with its current structure.

cmd/aws/create.go
```diff
@@ -161,7 +157,127 @@ func ValidateProvidedFlags(gitProvider string) error {
...
+	GetParameter(ctx context.Context, params *ssm.GetParameterInput, optFns ...func(*ssm.Options)) (*ssm.GetParameterOutput, error)
+}
+
+func GetLatestAMIFromSSM(ctx context.Context, ssmClient ssmClienter, parameterName string) (string, error) {
+	input := &ssm.GetParameterInput{
+		Name: aws.String(parameterName),
+	}
+	output, err := ssmClient.GetParameter(ctx, input)
+	if err != nil {
+		return "", fmt.Errorf("failed to initialise ssm client: %w", err)
```
patrickdappollonio, line 237: Need to decide on what english we use across the app, `initialize` vs `initialise`.

cmd/aws/create_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+			creds, err := getSessionCredentials(context.Background(), mockProvider)
+			if tt.expectedErr != nil {
+				assert.Nil(t, creds)
+				assert.EqualError(t, err, tt.expectedErr.Error())
+			} else {
+				assert.NotNil(t, creds)
+				assert.NoError(t, err)
+				assert.Equal(t, tt.creds, *creds)
+			}
```
patrickdappollonio, line 66: Small word of advice: `assert` in `stretchr/testify/assert` fails the test, but doesn't stop it.

If you fail the assertion, and then say you need the value being not nil, this will panic your test for the wrong reason.

You almost always will want to use `stretchr/testify/assert/require` with `require.Nil`, `require.EqualError()` and `require.Equal()` vs their `assert.X` counterparts.

cmd/aws/create_test.go
```diff
@@ -0,0 +1,402 @@
...
+			expectedErr: errors.New("failed to retrieve AWS credentials: failed to retrieve credentials"),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockProvider := &mockCredentialsProvider{
+				creds: tt.creds,
+				err:   tt.err,
+			}
```
patrickdappollonio, line 56: I see your intention here, but it's more recommendable to change the behaviour of the mock to dynamically set the function rather than its return values. 

I would recommend you to change the mock structure above to be:

```go
type mockCredentialsProvider struct {
	fnRetrieve (ctx context.Context) (aws.Credentials, error)
}

func (m *mockCredentialsProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
	if m.fnRetrieve == nil {
		return errors.New("not implemented")
	}

	return m.fnRetrieve(ctx)
}
```

Then each test can decide what `Retrieve()` looks like and return, rather than dynamically handling its error messages and returned values. 

The other reason is that this makes the mock usable just once before reinitialization if you want to test in the same test multiple case scenarios. You often don't want to go that way.

cmd/aws/create_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+			creds, err := getSessionCredentials(context.Background(), mockProvider)
+			if tt.expectedErr != nil {
+				assert.Nil(t, creds)
+				assert.EqualError(t, err, tt.expectedErr.Error())
```
patrickdappollonio, line 61: This line is also a byproduct of the issue above where each unit test cannot configure the function, but rather, the "expected error message".

This is also a very, very bad idea because it will be comparing strings to strings: if someone modifies your code and changes just the casing of the string (from lowercase to uppercase for example), now none of your tests will pass.

cmd/aws/create_test.go
```diff
@@ -0,0 +1,402 @@
...
+			},
+		}, nil
+	} else if m.architecture == "" {
+		return &ec2.DescribeImagesOutput{
+			Images: m.images,
+		}, nil
+	} else {
+		return nil, errors.New("unexpected architecture")
+	}
+}
```
patrickdappollonio, line 309: Same note as the mock that's at the top of the page. 

You might also want to put the mocks together near the tests that uses them or together at the end of the file, but not half and half where there's some at the top and some at the bottom.

cmd/aws/create_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+// func TestValidateAMIType(t *testing.T) {
```
patrickdappollonio, line 311: Comments should always be left out 😆 or moved to a branch for testing (and don't forget you can start branches off your own branch, doesn't have to be from main).

patrickdappollonio (changes requested):
+1 to Nathan's comments as well.

cmd/aws/create.go
jokestax: using initialize

cmd/aws/create_test.go
jokestax: got it thank you so much, it was `"github.com/stretchr/testify/require"`

cmd/aws/create.go
```diff
@@ -161,7 +159,116 @@ func ValidateProvidedFlags(gitProvider string) error {
 		log.Info().Msgf("%q %s", "gitlab.com", key.Type())
 	}
 
+	ssmClient := ssm.NewFromConfig(cfg)
+	ec2Client := ec2.NewFromConfig(cfg)
+	paginator := ec2.NewDescribeInstanceTypesPaginator(ec2Client, &ec2.DescribeInstanceTypesInput{})
+
+	if err := validateAMIType(ctx, amiType, nodeType, ssmClient, ec2Client, paginator); err != nil {
+		progress.Error(err.Error())
+		return fmt.Errorf("failed to validte ami type for node group: %w", err)
```
nathan-nicholson, line 168:
```suggestion
		return fmt.Errorf("failed to validate ami type for node group: %w", err)
```

nathan-nicholson (dismissed):
While it may make sense to split out much of this into a struct responsible, we can do that as part of another PR. :shipit:

cmd/aws/create_test.go
```diff
@@ -0,0 +1,417 @@
...
+			},
+		}, nil
+	} else if m.architecture == "" {
+		return &ec2.DescribeImagesOutput{
+			Images: m.images,
+		}, nil
+	} else {
+		return nil, errors.New("unexpected architecture")
+	}
+}
```
patrickdappollonio, line 97: These are still using the old fashion approach where instead of passing the function, you pass the returned values, making them usable exactly once before you have to reinstantiate them again.

cmd/aws/create_test.go
```diff
@@ -0,0 +1,417 @@
...
+			mockProvider := &mockCredentialsProvider{
+				fnRetrieve: func(ctx context.Context) (*aws.Credentials, error) {
+					return &tt.creds, tt.err
+				},
+			}
+
+			creds, err := getSessionCredentials(context.Background(), mockProvider)
+			if tt.expectedErr != nil {
+				require.Nil(t, creds)
+				require.EqualError(t, err, tt.expectedErr.Error())
```
patrickdappollonio, line 135: I mentioned in the previous review but it seems it was marked as resolved without actually resolving it. This is a very weak validation that depends on the fact that a) both haveto be error types, and b) both error string versions have to match. Not sure you want to do that for the reasons I provided before.

The same applies for the rest of the checks in the unit tests below.

cmd/root.go
```diff
@@ -36,6 +36,7 @@ var rootCmd = &cobra.Command{
 		fmt.Println("  kubefirst help")
 		progress.Progress.Quit()
 	},
+	SilenceErrors: true,
```
patrickdappollonio, line 39: ... how come?

patrickdappollonio (changes requested):

cmd/root.go
nathan-nicholson: Without this, we are doing a double print of errors, which looked pretty clunky. My understanding of this setting is that we're (aspirationally 😄) managing the printing of error messages so there was no need to tack it onto the end as well.

cmd/aws/create_test.go
jokestax: I am unclear and would need your help ig 😅

nathan-nicholson: What Patrick is saying is that you can define a mock implementation of DescribeImages to the mock struct instead of using this single definition. You can see an example of this on the mockCredentialsProvider. It's a function with the same signature as the Method, but that can be dynamically defined for each test case.

patrickdappollonio: Yes, sorry, but exactly what Nathan said. I explained it in detail in one of my previous comments but I think because you edited the file either Github or you marked the issue as "resolved", so the comment is collapsed now.

cmd/root.go
patrickdappollonio: Oh right, we changed the `main()` function to `Fprintln(os.Stderr, ...)`, right? Got it. Thank you!

nathan-nicholson:
I didn't finish completely replacing the mocks, but I'm calling it for today. Spent more time fighting table tests than I'm proud to admit. May have been easier to just do individual test cases for most of these 🤷

cmd/aws/create.go
```diff
@@ -41,7 +46,14 @@ func createAws(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("invalid catalog apps: %w", err)
 	}
 
-	err = ValidateProvidedFlags(cliFlags.GitProvider)
+	ctx := context.Background()
```
patrickdappollonio, line 49:
```suggestion
	ctx := cmd.Context()
```

https://pkg.go.dev/github.com/spf13/cobra#Command.Context

cmd/aws/create.go
```diff
@@ -161,7 +159,120 @@ func ValidateProvidedFlags(gitProvider string) error {
...
+
+func getSessionCredentials(ctx context.Context, cfg aws.CredentialsProvider) (*aws.Credentials, error) {
+	// Retrieve credentials
+	creds, err := cfg.Retrieve(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve AWS credentials: %w", err)
+	}
+
+	return &creds, nil
+}
```
patrickdappollonio, line 184:
```suggestion
func getSessionCredentials(ctx context.Context, cp aws.CredentialsProvider) (*aws.Credentials, error) {
	// Retrieve credentials
	creds, err := cp.Retrieve(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to retrieve AWS credentials: %w", err)
	}

	return &creds, nil
}
```

cmd/aws/create.go
```diff
@@ -161,7 +159,120 @@ func ValidateProvidedFlags(gitProvider string) error {
...
+
+func validateAMIType(ctx context.Context, amiType, nodeType string, ssmClient ssmClienter, ec2Client ec2Clienter, paginator paginator) error {
+	ssmParameterName, ok := supportedAMITypes[amiType]
+	if !ok {
+		return fmt.Errorf("not a valid ami type: %q", amiType)
+	}
+
+	log.Info().Msgf("ami type is  %s", amiType)
```
patrickdappollonio, line 192: Do we need this logged?

```suggestion
```

go.mod
```diff
@@ -26,138 +28,163 @@ require (
...
+	go.opentelemetry.io/contrib/detectors/gcp v1.29.0 // indirect
+	go.opentelemetry.io/otel/sdk v1.32.0 // indirect
+	go.opentelemetry.io/otel/sdk/metric v1.29.0 // indirect
+	go.starlark.net v0.0.0-20230525235612-a134d8f9ddca // indirect
+	google.golang.org/grpc/stats/opentelemetry v0.0.0-20240907200651-3ffb98b2c93a // indirect
+	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
+	sigs.k8s.io/kustomize/kustomize/v5 v5.4.2 // indirect
+)
+
+require (
```
patrickdappollonio, line 78:
```suggestion
```

And perhaps do another `go mod tidy`.

go.mod
```diff
@@ -232,124 +257,124 @@ require (
...
+	k8s.io/kubectl => k8s.io/kubectl v0.31.0
+	k8s.io/kubelet => k8s.io/kubelet v0.31.0
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.31.0
+	k8s.io/metrics => k8s.io/metrics v0.31.0
+	k8s.io/mount-utils => k8s.io/mount-utils v0.31.0
+	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.31.0
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.31.0
+	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.31.0
+	k8s.io/sample-controller => k8s.io/sample-controller v0.31.0
+)
```
patrickdappollonio, line 372: Bumping this version might have some unintended consequences. If the bump isn't needed, I would advise not to migrate up until all the supported clusters are tested (specifically, Kubernetes clusters running <= 1.31)

patrickdappollonio: If you can try 1.28, I'll be happy to approve.

patrickdappollonio (dismissed):

go.mod
jokestax: tested ✅

nathan-nicholson (approved):
:shipit:

PR #2360, opened 2025-01-13
housekeeping: fix catalog apps generation and validation.

patrickdappollonio:
## Description

Doing the AWS STS work I found this function that seems to be buggy. Not only loading information into each catalog app wasn't working because it was operating on a copy of the original value, but the error message wasn't being handled at all.

The function returned 2 important values, a boolean that would say if the catalog was incorrect and an error that would say if the catalog was incorrect, thus making it `if err != nil` to always have a non-true `bool` return.

If the callers were handling this in a way that would require the "bug" to be there then this would break the current flow.

## Related Issue(s)

Fixes #


internal/catalog/catalog.go
```diff
@@ -36,84 +35,83 @@ func NewGitHub() *git.Client {
...
 
-func ValidateCatalogApps(catalogApps string) (bool, []apiTypes.GitopsCatalogApp, error) {
-	items := strings.Split(catalogApps, ",")
-
-	gitopsCatalogapps := []apiTypes.GitopsCatalogApp{}
+func ValidateCatalogApps(catalogApps string) ([]apiTypes.GitopsCatalogApp, error) {
 	if catalogApps == "" {
-		return true, gitopsCatalogapps, nil
+		// No catalog apps to install
+		return nil, nil
```
nathan-nicholson, line 66: Previously we were returning an empty slice, but now we are returning nil. Will downstream consumers of this be able to handle that?

nathan-nicholson (approved):

PR #2364, opened 2025-01-15
Generate app scaffold from CLI

mrsimonemms:
## Description

Adds the `generate app-scaffold` command. Tried to make this as generic as possible because the users are unlikely to have a `values.yaml` file that matches Metaphor

## Related Issue(s)

Fixes https://github.com/konstructio/kubefirst-api/issues/488

## How to test

Run the app:
```bash
go run . generate app-scaffold # will prompt for name
go run . generate app-scaffold -n app # will generate in ./registry
go run . generate app-scaffold -n app --output-path /tmp/my-app # will generate in /tmp/my-app
go run . generate app-scaffold -n app --environment hello,world # will only generate environments "hello" and "world"
```

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -7,6 +7,10 @@ See the LICENSE file for more details.
 package cmd
 
 import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/konstructio/kubefirst/internal/generate"
 	"github.com/konstructio/kubefirst/internal/progress"
 	"github.com/spf13/cobra"
 )
@@ -26,19 +30,27 @@ func GenerateCommand() *cobra.Command {
 func generateApp() *cobra.Command {
 	var name string
 	var environments []string
+	var outputPath string
 
 	appScaffoldCmd := &cobra.Command{
 		Use:              "app-scaffold",
 		Short:            "scaffold the gitops application repo",
 		TraverseChildren: true,
-		Run: func(_ *cobra.Command, _ []string) {
-			progress.Success("hello world")
+		RunE: func(_ *cobra.Command, _ []string) error {
+			if err := generate.AppScaffold(name, environments, outputPath); err != nil {
+				progress.Error(err.Error())
+				return fmt.Errorf("error scaffolding app: %w", err)
+			}
+
+			progress.Success(fmt.Sprintf("App successfully scaffolded: %s", name))
+			return nil
 		},
 	}
 
 	appScaffoldCmd.Flags().StringVarP(&name, "name", "n", "", "name of the app")
 	appScaffoldCmd.MarkFlagRequired("name")
 	appScaffoldCmd.Flags().StringSliceVar(&environments, "environments", []string{"development", "staging", "production"}, "environment names to create")
+	appScaffoldCmd.Flags().StringVar(&outputPath, "output-path", filepath.Join(".", "registry", "environments"), "location to save generated files")
 
 	return appScaffoldCmd
 }
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -1,0 +1,44 @@
+/*
+Copyright (C) 2021-2025, Kubefirst
+
+This program is licensed under MIT.
+See the LICENSE file for more details.
+*/
+package cmd
+
+import (
+	"github.com/konstructio/kubefirst/internal/progress"
+	"github.com/spf13/cobra"
+)
+
+func GenerateCommand() *cobra.Command {
+	generateCommand := &cobra.Command{
+		Use:   "generate",
+		Short: "code generator helpers",
+	}
+
+	// wire up new commands
+	generateCommand.AddCommand(generateApp())
+
+	return generateCommand
+}
+
+func generateApp() *cobra.Command {
+	var name string
+	var environments []string
+
+	appScaffoldCmd := &cobra.Command{
+		Use:              "app-scaffold",
+		Short:            "scaffold the gitops application repo",
+		TraverseChildren: true,
+		Run: func(_ *cobra.Command, _ []string) {
+			progress.Success("hello world")
+		},
+	}
+
+	appScaffoldCmd.Flags().StringVarP(&name, "name", "n", "", "name of the app")
+	appScaffoldCmd.MarkFlagRequired("name")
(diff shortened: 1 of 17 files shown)


cmd/generate.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright (C) 2021-2025, Kubefirst
```
mrsimonemms, line 2: Used Kubefirst for consistency with rest of the project - there's a task to update everything to `Konstruct Inc`, but outside of this work.

cmd/generate.go
```diff
@@ -0,0 +1,61 @@
...
+	var outputPath string
+
+	appScaffoldCmd := &cobra.Command{
+		Use:              "app-scaffold",
+		Short:            "scaffold the gitops application repo",
+		TraverseChildren: true,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			if err := generate.AppScaffold(name, environments, outputPath); err != nil {
+				progress.Error(err.Error())
+				return errors.Wrap(err, "error scaffolding app")
```
patrickdappollonio, line 44: There's no need to use an external package for error wrapping, it's now in the std library:

```suggestion
				return fmt.Errorf("error scaffolding app: %w")
```

cmd/generate.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+
+			progress.Success(fmt.Sprintf("App successfully scaffolded: %s", name))
+			return nil
+		},
+	}
+
+	cwd, err := os.Getwd()
+	cobra.CheckErr(err)
```
patrickdappollonio, line 53: You don't need to get the current working directory if the path to the flag starts with `./`.

The entire std library (with a few exceptions) can handle `./` whenever it's file opening and all that.

internal/generate/files.go
```diff
@@ -0,0 +1,37 @@
...
+
+	f.data[file] = content
+}
+
+func (f *Files) Save(filePrefix string) error {
+	for file, content := range f.data {
+		name := filepath.Join(filePrefix, file)
+
+		if err := os.MkdirAll(filepath.Dir(name), 0o755); err != nil {
+			return errors.Wrap(err, "failed to create directory")
```
patrickdappollonio, line 28: Same note about `errors.Wrap()`.

internal/generate/files.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+
+		if err := os.WriteFile(name, content.Bytes(), 0o644); err != nil {
+			return errors.Wrap(err, "failed to write file")
```
patrickdappollonio, line 32: Same note about `errors.Wrap()`.

internal/generate/scaffold.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		if err := files.Save(filepath.Join(outputPath, env)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func generateAppScaffoldEnvironmentFiles(appName, environment string) (*Files, error) {
```
patrickdappollonio, line 40: Would you mind writing a unit test for this? 

It seems there are a few things about this function below that might panic, like accessing boundaries without checks (`s[1:]`), stopping on a directory and more.

internal/generate/scaffold.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	files := &Files{}
+	err = fs.WalkDir(scaffoldFS, ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return errors.Wrap(err, "error walking directory")
+		}
+
+		if d.IsDir() {
+			return nil
```
patrickdappollonio, line 61: Might want to make sure you want to "skip" directories rather than "return".

There's this: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/io/fs/walk.go;l=12-15

internal/generate/scaffold.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		// Get the file name without the root path
+		s := strings.Split(path, string(filepath.Separator))
+		file := strings.Join(s[1:], string(filepath.Separator))
```
patrickdappollonio, line 66: Might be better to use something like https://pkg.go.dev/path/filepath#Base

patrickdappollonio (changes requested):

internal/generate/scaffold.go
mrsimonemms: This seems to be incorrect - it skips the directory's content which, as I want the nested files, breaks the desired functionality.

internal/generate/scaffold.go
mrsimonemms: They don't do quite the same thing. This takes the string `scaffold/{{ .AppName }}/values.yaml` and returns `{{ .AppName }}/values.yaml`.

For the same string, `filepath.Base` would return `values.yaml`.

I tried in vain to find a `filepath` function which removes just the root directory - if you know of something that does this, please share.

internal/generate/scaffold.go
mrsimonemms: I've added a golden files test. Originally, I'd considered the `panic` chances low because it comes from an embedded file system, but you're probably right.

This also checks the file content is generated correctly too.

internal/generate/scaffold.go
patrickdappollonio: Can you do a bounds check if none of the stdlib can handle it? That `s[1:]` could panic if the path is not big enough to be separated.

You can prove this with a unit test too btw.

internal/generate/scaffold.go
patrickdappollonio: Ah gotcha, so you want to find the directory still, just not process the directory itself but the contents in it.

internal/generate/scaffold.go
mrsimonemms: It's always going to be at least one because of how the `embed` works, but I've reworked it

internal/generate/scaffold.go
mrsimonemms: Correct

jokestax (approved):
thank you @mrsimonemms ❤️

PR #2366, opened 2025-01-16
feat: remove beta

jokestax:
## Description
remove beta tag for akamai and azure as both support physical cluster


mrsimonemms (changes requested):
Can you update the "in beta" section in `/README.md` please?

See #2368 for an example

mrsimonemms (approved):

PR #2423, opened 2026-01-31
feat: upgrade kubefirst api tag

CristhianF7:
- Upgrading kubefirst api v1.136.0


johndietz (approved):

johndietz:
thanks so much cris 🧡

